PR #38, opened 2016-05-12
SimulatedChopper checks overrides for states and transitions

### LINKED ISSUE #22: StateMachine should check that keywords in state dict are valid
The dict that overrides chopper states should not be allowed to contain new states that are not already in the chopper. The same goes for transitions - there should be a check that excludes transitions which would add new states.

MichaelWedel:
This closes #22.

Added a more strict update function for dictionaries that raises an exception when the update dictionary contains keys that are not in the base dictionary. Not entirely sure `utils` is the best name, maybe you have another suggestion?

There are also a few tests, but the one for `SimpleChopper` only has test cases for this specific issue.

### DIFF
--- a/simulation/core/utils.py
+++ b/simulation/core/utils.py
@@ -28,11 +28,10 @@ def dict_strict_update(base_dict, update_dict):
     :param base_dict: The dict that is to be updated. This dict is modified.
     :param update_dict: The dict containing the new values.
     """
-    if update_dict is not None:
-        additional_keys = update_dict.viewkeys() - base_dict.viewkeys()
-        if len(additional_keys) > 0:
-            raise RuntimeError(
-                'The update dictionary contains keys that are not part of the base dictionary: {}'.format(
-                    str(additional_keys)))
+    additional_keys = update_dict.viewkeys() - base_dict.viewkeys()
+    if len(additional_keys) > 0:
+        raise RuntimeError(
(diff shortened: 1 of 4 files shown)


simulation/core/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    This function updates base_dict with update_dict if and only if update_dict does not contain
+    keys that are not already in base_dict. It is essentially a more strict interpretation of the
+    term "updating" the dict.
+
+    If update_dict contains keys that are not in base_dict, a RuntimeError is raised.
+
+    :param base_dict: The dict that is to be updated. This dict is modified.
+    :param update_dict: The dict containing the new values.
+    """
+    if update_dict is not None:
```
MikeHart85: Is this to allow something like...

``` python
foo = {...}
bar = None
dict_strict_update(foo, bar)
```

.. to pass without error (and without doing anything)?

If so, the checks against `None` in the SimChopper `__init__` should probably be removed?

Although... I think my gut feeling is that passing in a `None` value should throw whatever error it causes (same as any other not-dict-like value). Allowing it when the normal `update` doesn't, doesn't feel very strict.

MichaelWedel: You're probably right it is a bit strange to check for `None` and not for any other types that are incompatible. I'll simply remove the line.

PR #40, opened 2016-05-13
Setup basic continuous integration using travis-ci

### LINKED ISSUE #34: Setup build servers
As we're getting more unit tests, we should actually use them and start looking into setting up a Jenkins job that runs the unit tests when a PR is created. I suppose this is something where the DST-group may be able to help us.

Are there other things we want to check? I could imagine putting our documentation onto https://readthedocs.org/, maybe that's something that could be run nightly?

MichaelWedel:
This closes #34.

For the time being this solution should work for us. At least our tests are run each time a PR is opened and each time something is merged to master. [Travis-CI](https://travis-ci.org/) also seems to support deployment to docker, so this is actually quite convenient.

I'll open another issue to evaluate this after a while and see whether we want to keep it like this or switch to a Jenkins solution.

@Jon-Taylor, @OwenArnold, what do you think about this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,7 @@ python:
 install:
   - pip install -r requirements-dev.txt
 script: nosetests
+
+branches:
+  only:
+    - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: python
+python:
+  - "2.7"
+install:
+  - pip install -r requirements-dev.txt
+script: nosetests
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/DMSC-Instrument-Data/plankton.svg?branch=master)](https://travis-ci.org/DMSC-Instrument-Data/plankton)
+
 # Plankton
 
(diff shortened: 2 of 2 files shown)


OwenArnold:
Sounds good to me. This is probably the lightest approach to getting CI running against this code base.

OwenArnold:
@MichaelWedel. Yes this is working nicely. I think we ought to have some badges for our CI fed back. See changes on my [fork](https://github.com/OwenArnold/plankton/commit/6a83aa638e0e447443902a852349f69219400d69)

Also. IMHO, I think running this on all branches would be a better option than just master.

MichaelWedel:
It's running on master and pull requests. I thought on the normal development branches it would be enough to run the tests locally, otherwise a lot of e-mails are sent.

I was going to add a badge once this is merged so that the badge actually shows something. But I suppose we could add it to this PR.

MichaelWedel:
Okay I added the badge. Right now it's showing `unknown` but I suppose that goes away once master has been built once.

OwenArnold:
@MichaelWedel Yes it will go away.

PR #45, opened 2016-05-20
Update StateMachine documentation

### LINKED ISSUE #43: Update documentation for StateMachine
The StateMachine class and module received quite a few updates and overhauls during out recent visit to the DMSC.

We should update and add documentation for all the things we changed and added.

MikeHart85:
Closes #43.

StateMachine documentation was well out of date since the development burst during our visit.

Added documentation to new classes, updated old documentation for changes made, improved terminology and wording in some places.

### DIFF
--- a/simulation/chopper/device.py
+++ b/simulation/chopper/device.py
@@ -162,7 +162,6 @@ class SimulatedChopper(CanProcessComposite, object):
             'states': state_handlers,
             'transitions': transition_handlers,
         }, context=self._context)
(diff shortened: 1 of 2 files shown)


MikeHart85:
Also changed order of parameters in `bind_handlers_by_name`. Seems to make more sense this way around, should be more likely to specify override than change the default prefixes. 

And removed the call to this function from SimulatedChopper. Since we have state handlers in separate State classes there, not in the SimulatedChopper class itself, there's no need to call it.

PR #67, opened 2016-06-30
Added automatic parking switch

### LINKED ISSUE #60: Add automatic parking switch
Like in the chic, automatic parking on stop must be implemented. There should be a boolean property that can be set. If this is active, the chopper will move the disc into the parked state automatically, when the stopped-state is reached. Otherwise it will remain in the stopped state. This needs to be exposed via an additional PV.

MichaelWedel:
Closes #60.

There is now a new switch in the chopper context that will automatically trigger `park_commanded` when `stopped` is entered if activated. It's exposed via a new PV that is an enum (`false`, `true`) - this makes the usual EPICS 0, 1 syntax possible as well as using `false` and `true`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ At a minimum, the following PVs will be exposed:
 | Phs  |  Phase setpoint. | Degree | Read/Write |
 | ParkAng-RB  |  Current position of chopper disc if in parked state. | Degree | Read |
 | ParkAng  |  Position to which the disc should rotate in parked state. | Degree | Read/Write |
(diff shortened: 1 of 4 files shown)


MichaelWedel:
I'm not so worried about decreased test coverage, there are no tests yet for the chopper. But it is a good opportunity to create a ticket to add some.

MichaelWedel:
I've adjusted the settings for coveralls so that it's not super sensitive. But the docker hub issue is still there.

MichaelWedel:
It's working now. The last image is from master now and PRs are not dockerized.

Issue #71, opened 2016-08-04
Adapters should have their own ArgumentParser


MichaelWedel:
Instead of having the `--parameters` parameter, the simulation script should forward additional parameters to the adapter, which then gets its own ArgumentParser, as described in #70.

This is necessary because the script fails if `--parameters` is omitted, which is unacceptable.

Issue #70, opened 2016-07-15
Startup Consistency Issues


MikeHart85:
There are a few consistency issues with the current startup process that I think should be addressed.

Some of these could be addressed separately, others are perhaps better done as a single PR.
### Adapter vs Protocol

We have an "adapters" folder which is to contain implementations for a variety of protocols. For example it contains an "epics.py" to implement the EPICS adapter.

But the startup arguments refer to "epics" as a "protocol", not an adapter. Even though the input to that "protocol" parameter is used to load "epics.py" in the "adapters" folder.

Additionally, it also introduces an "adapter" parameter that refers to the name of the class defined inside the adapter module. For example, "EpicsAdapter", which is defined in "epics.py". Do we really need to expose this? Will we ever define multiple adapters in a single adapter file? What benefit would this have over one adapter per "adapters/*.py" file?
### `import_adapter` vs other `import_*`s

`import_adapter` returns a Class reference, that then needs to be instantiated.

The others imports return an instance directly.

The `import_*`s and `get_available_submodules` are also free-standing functions that are scattered across various `__init__.py` files and `utils.py`. I think it would be better to group them together, if possible, perhaps in a central factory class as part of core.
### ArgumentParser vs Ad Hoc `--parameters`

Plankton parameters are parsed using ArgumentParser, which provides a standard `-x "short version"` / `--thisisthe "long version"` parameter interface to the user, along with friendly usage and error reporting.

Parameters to the adapter, however, have to be passed via a "--parameters" parameter, and in a "key=value" format... and, naturally, are parsed in a completely different way. Which leads to any incorrect input throwing ugly stack traces instead of friendly error reports.

I think a better solution would be to have the ArgumentParser in `simulation.py` parse what it can, and pass everything left over along to the adapter exactly as it is. The adapter should then use an ArgumentParser of its own to parse the rest. To explicitly split plankton parameters from adapter parameters, a `--` could be used.

For example:
`$ python simulation.py -d chopper -p epics --prefix SIM:`

Or, using a `--` separator and having the PV Prefix as a positional argument instead:
`$ python simulation.py -d chopper -p epics -- SIM:`
### Should be able to execute `simulation.py`

Always having to pass it to python is cumbersome and unnecessary.
### Device vs Setup

A device cannot be started without a setup or vice versa.

Specifying a device without a setup, we can reasonable default to the "default" setup.

[...]

Issue #56, opened 2016-06-16
Integration with IBEX


MikeHart85:
Depends on #53.

As a proof-of-concept, we want to implement a simulator of a simple device used in IBEX and configure it so that IBEX can interact with that simulation in the same way as it would if it was a real device.

This likely requires a TCP stream adapter.

MikeHart85:
This is largely done via #72.

To connect IBEX to a plankton simulated device, `st.cmd` in the IOCs `iocBoot` directory should be modified to remove or comment out this section:

```
drvAsynSerialPortConfigure("$(DEVICE)", "$(PORT)", 0, 0, 0, 0)
asynSetOption("$(DEVICE)", -1, "baud", "$(BAUD=9600)")
asynSetOption("$(DEVICE)", -1, "bits", "$(BITS=8)")
asynSetOption("$(DEVICE)", -1, "parity", "$(PARITY=none)")
asynSetOption("$(DEVICE)", -1, "stop", "$(STOP=1)")
asynOctetSetInputEos("$(DEVICE)", -1, "$(OEOS=\n)")
```

And replace it with this:

```
drvAsynIPPortConfigure("$(DEVICE)", "192.168.99.100:9999")
```

Where the IP and port is adjusted as required. If running on Windows, and using the Plankton Docker image, it should be launched with a `-p 9999:9999` (or whatever port number is being used) to forward the port from the container to the VM's network adapter.

Going to keep this open for a bit, probably until we hand off the Linkam to the IBEX team.

MichaelWedel:
This has been achieved with the addition of the Linkam, so I'm closing this issue.

PR #72, opened 2016-08-04
Linkam T95 Device with TCP Stream Adapter

### LINKED ISSUE #53: Expose parts of a simulated device through different protocols
The TDC events may require streaming via some protocol, while the rest needs to be controlled via EPICS. Details follow.

This is linked to #39.

MikeHart85:
Closes #53 and #56.

This PR implements a Linkam T95 Temperature Controller. Behaviour is still very rudimentary and some features aren't fully implemented. But basic behaviour is provided and IBEX can interact with it sufficiently to serve as a basic for further development to refine behaviour and fill in edge cases.

This PR also implements a TCP Stream adapter to allow the Linkam to communicate with the IOC used by IBEX. Currently the port it listens on is fixed to 9999. This should be addressed once some other issues with the startup process are resolved (see #70, #71).

Unit tests should be added in a separate issue.

### DIFF
--- a/adapters/stream.py
+++ b/adapters/stream.py
@@ -17,55 +17,62 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # *********************************************************************
 
-import SocketServer
-import re
+
+import asyncore
+import asynchat
+import socket
 
 from adapters import Adapter
 from datetime import datetime
 
 
-class StreamHandler(SocketServer.StreamRequestHandler):
-    def handle(self):
-        request = self.rfile.readline().strip()
-        reply = None
+class StreamHandler(asynchat.async_chat):
+    def __init__(self, sock, target, bindings):
+        asynchat.async_chat.__init__(self, sock=sock)
+        self.set_terminator(bindings['meta']['in_terminator'])
+        self.target = target
+        self.bindings = bindings
+        self.buffer = []
 
-        for command, funcname in self.server.bindings.iteritems():
+    def collect_incoming_data(self, data):
+        self.buffer.append(data)
+
+    def found_terminator(self):
+        request = ''.join(self.buffer)
+        reply = None
+        self.buffer = []
+
+        for command, funcname in self.bindings['commands'].iteritems():
             if request.startswith(command):
-                func = getattr(self.server.target, funcname)
+                func = getattr(self.target, funcname)
                 args = request[len(command):]
                 reply = func(args) if args else func()
 
-        """
-        if request == "STATE?":
-            reply = self.server.target.state + "\n"
-        if request == "INIT!":
-            self.server.target.initialize()
-            reply = "Initializing\n"
-        if request == "STATS?":
-            speed = self.server.target.speed
-            phase = self.server.target.phase
-            reply = "%.2f rpm @ %.2f deg\n" % (speed, phase)
-        if request.startswith("GO!"):
-            m = re.match("GO!(\d+(?:\.\d+)?)@(\d+(?:\.\d+)?)", request)
-            if m is None:
-                reply = "INVALID FORMAT!\n"
-            else:
-                self.server.target.targetSpeed = float(m.group(1))
-                self.server.target.targetPhase = float(m.group(2))
-                self.server.target.start()
-                reply = "START COMMANDED!\n"
-        """
-
         if reply is not None:
-            self.request.sendall(str(reply))
+            self.push(str(reply) + self.bindings['meta']['out_terminator'])
+
+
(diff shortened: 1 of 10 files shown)


adapters/stream.py
```diff
@@ -0,0 +1,96 @@
...
+        reply = None
+        self.buffer = []
+
+        for command, funcname in self.bindings['commands'].iteritems():
+            if request.startswith(command):
+                func = getattr(self.target, funcname)
+                args = request[len(command):]
+                try:
+                    reply = func(args) if args else func()
+                except Exception:
```
MichaelWedel, line 51: Maybe in this case there should be a comment why the exception is eaten. Normally I wouldn't consider that necessary but since there is this very specific reason I think it should be documented.

adapters/stream.py
```diff
@@ -0,0 +1,96 @@
...
+                    pass
+
+        if reply is not None:
+            self.push(str(reply) + self.bindings['meta']['out_terminator'])
+
+
+class StreamServer(asyncore.dispatcher):
+    def __init__(self, host, port, target, bindings):
+        asyncore.dispatcher.__init__(self)
```
MichaelWedel, line 62: For consistency, maybe this could be transformed to the `super(StreamServer, self).__init__()`-syntax.

core/utils.py
```diff
@@ -34,6 +34,7 @@ def get_available_submodules(package, search_path=None):
     file_name, path, descriptor = imp.find_module(package, search_path)
 
     submodule_candidates = [extract_module_name(osp.join(path, entry)) for entry in listdir(path)]
+    submodule_candidates = filter(lambda x: x is not None, submodule_candidates)
 
     return [submodule for submodule in submodule_candidates if is_module(submodule, [path])]
```
MichaelWedel, line 38: I think the filter could be added to the list comprehension's `if`-statement: `... if submodule is not None and is_module(submodule, [path])`. But I think the `ValueError`-exception in `is_module` catches this case, there are some tests where `extract_module_name` returns `None`, so that `is_module` returns false.

adapters/stream.py
```diff
@@ -0,0 +1,96 @@
...
+        pair = self.accept()
+        if pair is not None:
+            sock, addr = pair
+            print "Client connect from %s" % repr(addr)
+            StreamHandler(sock, self.target, self.bindings)
+
+
+class StreamAdapter(Adapter):
+    def run(self, target, bindings, *args, **kwargs):
+        StreamServer("0.0.0.0", 9999, target, bindings)
```
MichaelWedel, line 80: As you wrote it should be possible to change host and port, especially to make this more restrictive, but that is a separate issue to be solved after #71.

adapters/stream.py
MikeHart85, line 51: Good idea, I'll add one.

devices/linkam_t95/device.py
```diff
@@ -0,0 +1,183 @@
...
+        # TODO: Is not having leading zeroes / 4 digits an error?
+        # TODO: What is the upper limit in the real device?
+        rate = int(param)
+        if 1 <= rate <= 9999:
+            self._context.temperature_rate = rate / 100.0
+        print "New rate: %.2f C/min" % (self._context.temperature_rate,)
+        return ""
+
+    def setLimit(self, param):
+        # TODO: Is not having leading zeroes / 4 digits an error?
```
MichaelWedel, line 141: From the values below it seems this should be 5 characters.

adapters/stream.py
MikeHart85, line 62: Unfortunately, these are old style classes and so we have to use the old syntax.

core/utils.py
MikeHart85, line 38: As discussed, I'll remove that line and get rid of the errant print statement in is_module instead.

devices/linkam_t95/device.py
MikeHart85, line 141: Well, I'm not so certain one way or the other...

The -1960 to 15000 values are from the documented limits on the temperature _readout_. They're not necessarily correct for the input. The documentation for the input doesn't specify limits or whether there should be leading zeroes. It only provides an example setting the rate to 2000, and the limit to 1250 and -1250.

The .proto file used by IBEX, on the other hand, uses sprintf style formatting that implies four digits (not counting the `-` sign) and leading zeroes. No indication for upper or lower limits there either.

adapters/stream.py
MikeHart85, line 80: Yeah, I don't like how it binds `0.0.0.0` either. We should fix that soon. But, as you say, we need to deal with parameter handling first.

PR #73, opened 2016-08-09
ArgumentParsers for Adapters

### LINKED ISSUE #71: Adapters should have their own ArgumentParser

MikeHart85:
Closes #71.

This PR adds `ArgumentParser`s to the adapters and modifies startup such that general plankton arguments are handled in `simulation.py` and the remainders is passed on to the adapter. 

Plankton arguments should be separated from adapter arguments by a `--`.

Generally, Plankton should now be invoked as follows:

```
# docker run -it dmscid/plankton -d device_name -p {epics|stream} [-- [adapter arg] [...]]
```

Stream Adapter: Added `-p/--port` and `-b/--bind-address` arguments. Port defaults to 9999 and address to `0.0.0.0` (listen on all network adapters).

EPICS Adapter: The previous `--parameters pv_prefix=SIM:` now becomes `-- -p SIM:` or `-- --prefix SIM:`. It defaults to an empty string.

Examples:

```
# docker run -it dmscid/plankton -d chopper -p epics
# docker run -it dmscid/plankton -d chopper -p epics -- -p SIM1:
# docker run -it dmscid/plankton -d chopper -p epics -- --prefix SIM2:
# docker run -it dmscid/plankton -d linkam_t95 -p stream
# docker run -it dmscid/plankton -d linkam_t95 -p stream -- --bind-address localhost
# docker run -it -p 9999:1234 dmscid/plankton -d linkam_t95 -p stream -- -b 172.17.0.2 --port 1234
```

### DIFF
--- a/adapters/epics.py
+++ b/adapters/epics.py
@@ -74,9 +74,11 @@ class EpicsAdapter(Adapter):
         parser.add_argument('-p', '--prefix', help='Prefix to use for all PVs', default='')
         return parser.parse_args(arguments)
 
-    def run(self, target, bindings, prefix):
+    def run(self, target, bindings, arguments):
+        options = self._parseArguments(arguments)
+
         server = SimpleServer()
-        server.createPV(prefix=prefix, pvdb=bindings)
+        server.createPV(prefix=options.prefix, pvdb=bindings)
         driver = PropertyExposingDriver(target=target, pv_dict=bindings)
 
         delta = 0.0  # Delta between cycles
--- a/adapters/epics.py
+++ b/adapters/epics.py
@@ -70,7 +70,6 @@ class PropertyExposingDriver(CanProcess, Driver):
 
 class EpicsAdapter(Adapter):
     def run(self, target, bindings, arguments):
-        print arguments
         parser = ArgumentParser(description="Adapter to expose a device via EPICS")
         parser.add_argument('-p', '--prefix', help='Prefix to use for all PVs', default='')
         arguments = parser.parse_args(arguments)
--- a/adapters/epics.py
+++ b/adapters/epics.py
@@ -18,6 +18,7 @@
 # *********************************************************************
 
 from datetime import datetime
+from argparse import ArgumentParser
 
 from pcaspy import Driver, SimpleServer
 
@@ -68,9 +69,14 @@ class PropertyExposingDriver(CanProcess, Driver):
 
 
 class EpicsAdapter(Adapter):
-    def run(self, target, bindings, *args, **kwargs):
+    def run(self, target, bindings, arguments):
+        print arguments
+        parser = ArgumentParser(description="Adapter to expose a device via EPICS")
+        parser.add_argument('-p', '--prefix', help='Prefix to use for all PVs', default='')
+        arguments = parser.parse_args(arguments)
+
         server = SimpleServer()
-        server.createPV(prefix=kwargs['pv_prefix'], pvdb=bindings)
+        server.createPV(prefix=arguments.prefix, pvdb=bindings)
         driver = PropertyExposingDriver(target=target, pv_dict=bindings)
 
         delta = 0.0  # Delta between cycles
--- a/adapters/epics.py
+++ b/adapters/epics.py
@@ -69,13 +69,14 @@ class PropertyExposingDriver(CanProcess, Driver):
 
 
 class EpicsAdapter(Adapter):
-    def run(self, target, bindings, arguments):
+    def _parseArguments(self, arguments):
         parser = ArgumentParser(description="Adapter to expose a device via EPICS")
         parser.add_argument('-p', '--prefix', help='Prefix to use for all PVs', default='')
-        arguments = parser.parse_args(arguments)
+        return parser.parse_args(arguments)
 
+    def run(self, target, bindings, prefix):
         server = SimpleServer()
-        server.createPV(prefix=arguments.prefix, pvdb=bindings)
+        server.createPV(prefix=prefix, pvdb=bindings)
         driver = PropertyExposingDriver(target=target, pv_dict=bindings)
 
         delta = 0.0  # Delta between cycles
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
 FROM dmscid/epics-pcaspy:latest
 
-COPY . /plankton
+COPY requirements*.txt /plankton/
 
 RUN pip install -r plankton/requirements.txt && \
     pip install -r plankton/requirements-dev.txt
+
+COPY . /plankton
 
 ENTRYPOINT ["/init.sh", "/plankton/simulation.py"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 FROM dmscid/epics-pcaspy:latest
 
+# Copying these separately from the rest of plankton allows the
+# pip install step to be cached until the requirements change.
+# Rebuilding is faster and doesn't require an internet connection.
 COPY requirements*.txt /plankton/
 
 RUN pip install -r plankton/requirements.txt && \
--- a/adapters/__init__.py
+++ b/adapters/__init__.py
@@ -18,10 +18,18 @@
 # *********************************************************************
 
 import importlib
+import argparse
 
 
 class Adapter(object):
+    def _parseArguments(self, arguments):
+        return argparse.Namespace()
+
     def run(self, target, bindings, arguments):
+        options = self._parseArguments(arguments)
+        self.doRun(target, bindings, **vars(options))
+
+    def doRun(self, target, bindings, **options):
         pass
 
 
--- a/adapters/__init__.py
+++ b/adapters/__init__.py
@@ -22,14 +22,7 @@ import argparse
 
 
 class Adapter(object):
-    def _parseArguments(self, arguments):
-        return argparse.Namespace()
-
     def run(self, target, bindings, arguments):
-        options = self._parseArguments(arguments)
-        self.doRun(target, bindings, **vars(options))
-
-    def doRun(self, target, bindings, **options):
         pass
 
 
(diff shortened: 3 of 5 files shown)


adapters/epics.py
```diff
@@ -68,9 +69,14 @@ def doProcess(self, dt):
 
 
 class EpicsAdapter(Adapter):
-    def run(self, target, bindings, *args, **kwargs):
+    def run(self, target, bindings, arguments):
+        print arguments
```
MichaelWedel: This should probably be removed otherwise we'll discover weird things later on as it happened last time :)

MikeHart85:
@MichaelWedel 

I think this should be fine like this, but have a look whether it makes sense like this to you.

An alternative implementation might be to use the [subparser feature of argparse](https://docs.python.org/2/library/argparse.html#sub-commands). If we use that approach, I think the adapters would need to build and pass an argparser back to `simulation.py`, which would then register it as a subparser. But, since you don't know which one you need before you start parsing, it would need to request a subparser from every available protocol even though only one would end up being used.

adapters/epics.py
MikeHart85: Oops! Nice catch!

MikeHart85:
Ah, also, in #70, I mentioned that the device could be a positional argument (so no `-d` or `--device`) because it is always required and essentially **the** argument to pass in, with everything else being extra. So:

```
# docker run -it dmscid/plankton chopper -p epics -- --prefix SIM:
```

This issue might be a good place to make that change if we want to go ahead with it.

What do you think?

MichaelWedel:
I agree with what you say about the problems with the subparser approach, so I think giving each adapter its own parser is fine.

What we could maybe do is let this become part of the `Adapter` interface and not do the parsing in the `run`-method. There could be an additional method `Adapter.parseArguments` that does that and assigns the necessary parameters to member variables (or simply stores the parser namespace).

MichaelWedel:
So the parseArguments method or whatever would do nothing in the base class and `run` would only have `target` and `bindings`.

MikeHart85:
Feels a bit like abusing member variables as local globals. A method called `parseArguments` having side-effects on the behaviour of `run` doesn't seem right/intuitive to me. Passing the arguments directly to run does seem to make sense, since you're saying "run this adapter with these arguments".

But, building and running the argparser in `run` itself does clog it up a bit. How about just factoring it out to a `parseArguments` that takes the argument array and returns the parsed namespace? So in `run`, it would look like:

``` python
def run(self, target, bindings, arguments):
    options = self.parseArguments(arguments)
    StreamServer(options.bind_address, options.port, target, bindings)
```

Technically, it could/should be a private method, since it's not part of the interface like this... just a helper function.

MichaelWedel:
What about passing the arguments on construction of the adapter? They never change during the lifetime of the adapter anyway. That somehow seems equivalent to passing arguments to a script.

It could store the options as `self.options`, which can then be used in `run`.

```
adapter = CommunicationAdapter(arguments.adapter_args)
adapter.run(device, bindings)
```

Looks alright to me :)

MichaelWedel:
And yes, the actual parser construction/parsing should indeed be in a private method as you wrote.

MikeHart85:
That makes sense if we view the adapter as an immutable thing that we configure on creation, and then `run` when we're ready.

But, in that case, shouldn't we also move the `device` and `bindings` to the constructor as well? What makes them different from the (other) command-line arguments? They also never change during the lifetime of the adapter... at least currently.

I'm a bit worried that it won't necessarily stay that way though. Or that we won't always know how we're going to run the adapter at the point where we create it. What is the benefit of locking yourself into certain way of `run`ning on construction?

We've already talked a bit about how it's an issue that, currently, the adapter is really more of a "driver"... it contains the main loop of the simulation. But that's a problem if we want to use multiple protocols on the same simulation at the same time.

For example, the CHIC is supposed to support both EPICS and Modbus. I'm not sure... do you know whether that's an either-or situation, or whether both can be used at the same time? David also mentioned in a meeting a while back that they're considering using both Modbus and TCP Stream for one of their devices (with certain device features exposed on one protocol, and other features exposed on the other).

Freddie also mentioned that it would be a nice-to-have feature to control "meta" aspects of the simulation (like the delta T rate) via EPICS, even when simulating a TCP Stream device.

To support things like that, we'll need to redesign the way the simulation is driven to be independent from adapters, and allow multiple adapters to be active at the same time -- potentially serving very different functionality / parts of the simulation. Maybe we'll also want to dynamically enable/disable adapters and/or change their arguments? 

I'm not sure yet how that will turn out... that's why I'm hesitant to treat the arguments as immutable or anything other than "`run` this adapter like this right now", at least until we're sure it will stay that way.

MichaelWedel:
I'm not sure we should let us guide by stuff that we haven't discussed/planned properly yet. But nevertheless you're right about `device` and `bindings` being special in that case. So probably the best way for now is indeed to not change the run-function but move the parser construction/parsing out of the method itself.

One last possibility I'd like to throw in is something like this:

```
class Adapter(object):
    def _parseArguments(self, arguments):
        return {}

    def run(self, device, bindings, arguments):
        options = self._parseArguments(arguments)
        self.doRun(device, bindings, **options)
```

The concrete adapters would rename their `run`-method to `doRun` and add keyword arguments with defaults to the signature to match the parser. Just an idea.

MikeHart85:
You're right, of course, it's not good to develop from a "everything might change and we don't know how yet" perspective because then it doesn't matter how we implement things since... everything might change and we don't know how yet.

I'm just trying to explain why I'm gravitating towards the simplest possible thing that will work for now, and why I'm hesitant to move away from it: I suspect everything in and around the adapters will change drastically quite soon, so I'm drawn towards minimal solutions.

Which, of course, **is** developing from a "everything might change" perspective... but that's where we're at. I think that just highlights the need to tackle these design changes ASAP.

MichaelWedel:
True, it might. But having nice functions now is also nice :)

MikeHart85:
I've added the `_parseArguments` to the `Adapter` base class as you suggested, along with some other changes. Should be ready to review again, let me know if anything needs tweaking!

MichaelWedel:
I agree also in that we need to tackle these issues. I think I start seeing what you meant earlier, there might be quite strong coupling between certain parts of Adapter and Device. That being sa(i)d, I think removing `StoreNameValuePairs` is very good, it was basically duplicating ArgumentParser.

I would like to do a few of the manual tests you suggested before merging, would you still like to do the change from optional to positional parameter for device? Or should we just merge and do that in another (very small) PR?

adapters/epics.py
```diff
@@ -68,9 +69,14 @@ def doProcess(self, dt):
 
 
 class EpicsAdapter(Adapter):
-    def run(self, target, bindings, *args, **kwargs):
+    def _parseArguments(self, arguments):
+        parser = ArgumentParser(description="Adapter to expose a device via EPICS")
+        parser.add_argument('-p', '--prefix', help='Prefix to use for all PVs', default='')
+        return parser.parse_args(arguments)
+
+    def run(self, target, bindings, prefix):
```
MichaelWedel: This should be `doRun` now. I noticed pycharm is complaining about changing signature with respect to the base class. Not sure what the best solution is in this case. Maybe not defining `doRun` in the base class and instead doing the `if hasattr(self, 'doRun'): ...`-procedure? Did not see that coming, but it somehow makes sense that a method overriding a base class method should have the same signature.

I kind of like not using `kwargs`.

MikeHart85: Yeah, we can just remove `doRun` in the base class... I only added it since otherwise pycharm complained it didn't exist. It's smart enough not to complain if you check using `hasattr` though... I'm impressed.

Agreed about `kwargs`, it always makes things a bit awkward. There's one place in `StateMachine` that I ended up using it as well, but I'm not sure if it's possible to get rid of it there.

Having tried it, I don't think I'm a big fan of the result of using the `run`/`doRun` indirection here though...

Before, we had a very simple base class that defined a common interface: a `run` method that takes three named arguments. When implementing an adapter, PyCharm will even fill in the argument names for you when you start writing it. Very straightforward.

Now, the base class defines a public `run` that you're _not_ supposed to override, a private `_parseArguments` that you _are_ supposed to override, and expects you to implement a `doRun` which it doesn't define. And the signature of `doRun`, including the _names_ that the arguments _must_ have, depends on the contents of your `_parseArguments`. 

And all we've gained is that we don't have to call `_parseArguments` (or something like it, and assuming we need it) manually, no? Is that really worth it?

MichaelWedel:
We could generate the parser automatically from the function signature of `doRun` and have that as the base implementation of `_parseArguments`. That way it can still be overridden, but wouldn't have to be, leaving only `doRun`.

MichaelWedel:
To me this `arguments`-parameter feels like an unprocessed version of `kwargs`, which is why I like it even less.

MichaelWedel:
The auto generate thing would be a bit too much for this scope especially in Python 2 as I realise when reading the documentation of `inspect`, so that's probably not going to happen.

You're right it's more complicated now, but in return `doRun` is only doing one thing, running something instead of running something and parsing parts of the arguments. I think having to define two methods to implement an interface is not overly crazy. With regards to the signature depending on the parser I'd say it's the other way around. You have a method with a signature (`doRun`, which is the important part of the Adapter) and so you have to provide a parser to get the arguments. A certain coupling is not to possible to avoid in such a case...the parameter names can be different from the names in the function signature by using the options that `add_argument` has if necessary.

After writing this I'd go even further and say that _parseArguments should not even create the parser, instead it should do this:

```
def _parseArguments(self, arguments):
    return self.argument_parser.parse_args()
```

Turning into something that doesn't need to be re-implemented. The implementations should then provide `self.argument_parser`, maybe create it as part of their constructor or something like that.

MikeHart85:
As discussed on Slack, I've reverted back to just factoring out the parsing to a private helper method. 

We'll revisit after/during the adapter design changes.

PR #77, opened 2016-08-11
Starting chopper with target speed 0 should stop it instead

### LINKED ISSUE #66: Chopper state should transition to 'stopped' when speed setpoint is 0
When the speed setpoint is 0, the chopper should transition to `stopped` instead of `phase_locking`.

MichaelWedel:
This fixes #66.

The behavior has been implemented so that the `start`-method invokes the `stop`-method of the device if the target speed is 0. There are different ways to do this, for example changing the condition that needs to be true for transitioning into the `stopping`-state, but that requires more changes. Also, with this implementation it doesn't matter if the stopping mechanism changes.

To test you could bring the chopper to an arbitrary speed and phase and then set the target to 0 and start it. Instead of going to `accelerating`, the state should be `stopping`.

One problem this highlights is the "remember all commands ever issued" behavior. If the chopper is already stopped, the `stop_commanded`-switch is still set (this is not strictly related to this issue or the way the solution is implemented, it was there before). I don't think this is how the CHIC works for this particular case, I'll check with Andres. I think the `stop`-method should contain a check `if self._csm.state != 'stopped'` or something like that.

### DIFF
--- a/devices/chopper/device.py
+++ b/devices/chopper/device.py
@@ -209,7 +209,10 @@ class SimulatedChopper(CanProcessComposite, object):
 
     # Accelerating stuff
     def start(self):
-        self._context.start_commanded = True
+        if self._context.target_speed > 0.0:
+            self._context.start_commanded = True
+        else:
+            self.stop()
 
     @property
     def started(self):


MikeHart85:
Change looks good to me, I'll go ahead and merge.

I wouldn't be surprised if the CHIC does behave this way as well. We know it also has the "remember commands" behaviour in other situations, right?

It makes sense that it would also just flag a stop command instead of a start command when speed is 0, exactly the way you implemented it... it is the simplest way to achieve this behaviour. Would be good to double check, of course.

If the CHIC doesn't behave like this in this particular case, I think I would put the `state != 'stopped'` check in the `else` clause here rather than in the `stop` method. If we issue an actual stop command while stopped, the CHIC does remember that, doesn't it?

PR #80, opened 2016-08-12
Make plankton compatible with Python 2 & 3

### LINKED ISSUE #75: Python 3 support
Given that Python 2 support is running out at some point not so far in the future I think we should try making plankton run in Python 3 as well, eventually deciding whether it's possible to abandon Python 2 entirely.

MichaelWedel:
This fixes #75.

The changes in this branch make plankton run in both Python 2 & 3. To run the tests on Ubuntu, you can run `nosetests` and `nosetests3` in the plankton directory. You could also start the two devices manually to check that they are still running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
 
 python:
   - "2.7"
+  - "3.5"
 install:
   - pip install -r requirements-dev.txt
   - pip install coveralls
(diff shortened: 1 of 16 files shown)


MikeHart85:
Looks good, everything still works. 

Going to leave the docker images using Python 2.7 for now, we can update them in a separate PR if/when necessary.

PR #82, opened 2016-08-17
Main README Update

### LINKED ISSUE #31: Confusion of concepts now in README
The README mainly seems geared to explain the chopper instance you created. It also starts out by describing the state machine. As plankton will be used for multiple devices, the documentation should be re-ordered. 

What do you think @MichaelWedel @MikeHart85 ?

MikeHart85:
Closes #31.

This PR updates the main README.md to provide more information about Plankton itself rather than specifically the simulated chopper.

Information about the chopper has been moved to `devices/chopper/README.md` instead. Other devices could follow suit and provide their own READMEs to document the particulars of what they are simulating and how to interact with it.

The main README now contains quite a bit of information about how the framework works and what it aims to achieve, though it is quite high-level. Eventually, we may want to refactor that again by providing even more details while also pulling it out into separate MDs, linking to them from the main README. But for now I think this will do to provide an overview of how Plankton works and how to use it.


README.md
```diff
@@ -3,148 +3,193 @@
...
-```
-$ git clone https://github.com/DMSC-Instrument-Data/plankton.git
-```
+We consider a detailed device simulation to be one that can communicate using the same protocol as the real device, and that can very closely approximate real device behaviour in terms of what is seen through this protocol. This includes gradual processes, side-effects and error conditions.
+
+The purpose of Plankton is to provide a common framework to facilitate the development of such simulators. The framework provides a common set of tools and abstracts away protocol adapters, which helps minimize code replication and allows the developer of a simulated device to focus on capturing device behaviour.
+
+Potential use cases for detailed device simulators include:
+
+- Developing and testing software that interfaces with the device when it is unavailable
```
MichaelWedel: Maybe this could be reworded a bit to "Replacing a physical device for developing and testing software that interfaces with the device" or something like that. The "it" seems a bit ambiguous.

README.md
```diff
@@ -3,148 +3,193 @@
...
+Using a simulation for the above has the added benefit that, unlike most real devices, a simulation may be sped up / fast-forwarded past any lengthy delays or processes that occur in the device.
+
+
+## Framework Details
+
+The Plankton framework is built around a cycle-based statemachine that drives the device simulation, and shared protocol adapters that separate the communication layer from the simulated device.
+
+#### Cycle-based
+
+By cycle-based we mean that all processing in the framework occurs during "heartbeat" simulation ticks that propagate calls to `process` methods throughout the simulation, along with a Delta T parameter that contains the time that has passed since the last tick. The device simulation is then responsible for updating its state based on how much time has passed and what input has been received during that time.
```
MichaelWedel: Is there a better way to write "Delta T", maybe unicode? In any case I think it should be a lower case t for time...when I see T, I see "temperature" even though it's explicitly mentioned directly afterwards.

README.md
MikeHart85: Good idea, it's not really clear... that point is tough to word in a way that immediately makes sense.

README.md
```diff
@@ -3,148 +3,193 @@
...
+
+On Linux, this means that `EPICS_CA_ADDR_LIST` must include this networks broadcast address:
+
+```
+$ export EPICS_CA_AUTO_ADDR_LIST=NO
+$ export EPICS_CA_ADDR_LIST=172.17.255.255
+$ export EPICS_CAS_INTF_ADDR_LIST=localhost
+``` 
 
+On Windows and OSX, the docker0 network is inside of a virtual machine. If we want to communicate with it, we need to use an EPICS Gateway to forward EPICS requests and responses for us. We provide an [EPICS Gateway Docker image](https://hub.docker.com/r/dmscid/epics-gateway/) that can be used to do this relatively easily (detailed instructions on the linked page).
```
MichaelWedel: Here I recommend switching "we" to "you".

MikeHart85: Yeah, that doesn't sound good. I think I'll reword to avoid "we" or "you" altogether.

The "We provide" can probably stay though.

README.md
MikeHart85: As discussed, Δt sounds good to me (with it spelled out in parentheses the first time it shows up).

PR #83, opened 2016-08-18
Linkam T95 Unit Tests

### LINKED ISSUE #74: Linkam T95 needs unit tests
The Linkam T95 was added without corresponding unit tests.

They should be added.

MikeHart85:
Closes #74.

This PR adds unit tests to test the Linkam T95 device at the class level.

### DIFF
--- a/test/test_SimulatedLinkamT95.py
+++ b/test/test_SimulatedLinkamT95.py
@@ -71,8 +71,8 @@ class TestSimulatedLinkamT95(unittest.TestCase):
         self.assertEqual(status_bytes[0], '\x10')           # Heating status set
         self.assertNotEqual(status_bytes[6:10], '01b8')     # Temp != 44.0 C
 
-        # Finish off heating
-        linkam.process(0.5)
+        # Finish off heating (and overshoot a bit)
+        linkam.process(5)
         status_bytes = linkam.getStatus()
         self.assertEqual(status_bytes[6:10], '01b8')    # Temp == 44.0 C
 
@@ -102,8 +102,8 @@ class TestSimulatedLinkamT95(unittest.TestCase):
         self.assertEqual(status_bytes[0], '\x20')           # Cooling status set
         self.assertNotEqual(status_bytes[6:10], '0028')     # Temp != 4.0 C
 
-        # Finish off cooling
-        linkam.process(0.5)
+        # Finish off cooling (and overshoot a bit)
+        linkam.process(5)
         status_bytes = linkam.getStatus()
         self.assertEqual(status_bytes[6:10], '0028')    # Temp == 4.0 C
 
--- a/test/test_SimulatedLinkamT95.py
+++ b/test/test_SimulatedLinkamT95.py
@@ -44,22 +44,80 @@ class TestSimulatedLinkamT95(unittest.TestCase):
         linkam = SimulatedLinkamT95()
         status_bytes = linkam.getStatus()
 
-        self.assertEqual(len(status_bytes), 10)
-        self.assertFalse('\x00' in status_bytes)
-        self.assertEqual(status_bytes[0], '\x01')
-        self.assertEqual(status_bytes[1], '\x80')
-        self.assertEqual(status_bytes[2], '\x80')
+        self.assertEqual(len(status_bytes), 10)     # Byte array should always be 10 bytes
+        self.assertFalse('\x00' in status_bytes)    # Byte array may not contain zeroes
+        self.assertEqual(status_bytes[0], '\x01')   # Status byte should be 1 on startup
+        self.assertEqual(status_bytes[1], '\x80')   # No error flags should be set on startup
+        self.assertEqual(status_bytes[2], '\x80')   # The pump should not be active on startup
+        self.assertEqual(status_bytes[6:10], '00f0')  # Starting temperature 24C
 
     def test_simple_heat(self):
-        pass
+        linkam = SimulatedLinkamT95()
+        linkam.process(0)  # Initialize
+
+        # Issue T command to get into stopped state
+        linkam.getStatus()
+        linkam.process(0)
+
+        # Set up to heat from 24.0 C to 44.0 C at 20.00 C/min
+        linkam.setRate('2000')
+        linkam.setLimit('440')
+        linkam.start()
+        linkam.process(0)
+
+        # Heat for almost a minute but not quite
+        linkam.process(59.5)
+        status_bytes = linkam.getStatus()
+        self.assertNotEqual(status_bytes[6:10], '01b8')  # Temp != 44.0 C
+
+        # Finish off heating
+        linkam.process(0.5)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[6:10], '01b8')  # Temp == 44.0 C
+
+        # Should hold now, so temperature should not change
+        linkam.process(10)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[0], '\x30')  # Auto-holding at limit
+        self.assertEqual(status_bytes[6:10], '01b8')  # Temp == 44.0 C
 
     def test_simple_cool(self):
-        pass
+        linkam = SimulatedLinkamT95()
+        linkam.process(0)  # Initialize
+
+        # Issue T command to get into stopped state
+        linkam.getStatus()
+        linkam.process(0)
+
+        # Set up to cool from 24.0 C to 4.0 C at 20.00 C/min
+        linkam.setRate('2000')
+        linkam.setLimit('40')
+        linkam.start()
+        linkam.process(0)
+
+        # Heat for almost a minute but not quite
+        linkam.process(59.5)
+        status_bytes = linkam.getStatus()
+        self.assertNotEqual(status_bytes[6:10], '0028')  # Temp != 4.0 C
+
+        # Finish off heating
+        linkam.process(0.5)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[6:10], '0028')  # Temp == 4.0 C
+
+        # Should hold now, so temperature should not change
+        linkam.process(10)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[0], '\x30')  # Auto-holding at limit
+        self.assertEqual(status_bytes[6:10], '0028')  # Temp == 4.0 C
 
     def test_stop_command(self):
         pass
 
     def test_hold_and_resume(self):
+        pass
+
+    def test_hold_while_at_limit(self):
         pass
 
     def test_pump_command(self):
--- a/test/test_SimulatedLinkamT95.py
+++ b/test/test_SimulatedLinkamT95.py
@@ -1,0 +1,66 @@
+#  -*- coding: utf-8 -*-
+# *********************************************************************
+# plankton - a library for creating hardware device simulators
+# Copyright (C) 2016 European Spallation Source ERIC
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


test/test_SimulatedLinkamT95.py
```diff
@@ -0,0 +1,236 @@
...
+        # Set up to heat from 24.0 C to 44.0 C at 20.00 C/min
+        linkam.setRate('2000')
+        linkam.setLimit('440')
+        linkam.start()
+        linkam.process(0)
+
+        # Process for some time and then stop
+        linkam.process(10)
+        linkam.stop()
+        linkam.process(0)
```
MichaelWedel: As an unrelated idea, having seen this so often in this PR, could it be an option to add a convenience method for `process(0)`? Something like `processEmpty()` or `processNone()` or `nullCycle()` or...maybe you have a better idea :)

test/test_SimulatedLinkamT95.py
```diff
@@ -0,0 +1,236 @@
...
+        linkam.process(15)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[6:10], '0028')  # Temp == 4.0 C
+
+        # Make sure transitions to auto-holding
+        linkam.process(0)
+        status_bytes = linkam.getStatus()
+        self.assertEqual(status_bytes[0], '\x30')  # Auto-holding at limit
+
+    def test_pump_command(self):
```
MichaelWedel, line 205: This doesn't really assert anything.

test/test_SimulatedLinkamT95.py
MikeHart85: Could just make 0 the default value for dt, no? 

If we make a special function for it, the name should be as intuitive as possible to ensure its understood what it actually does. Especially since we're not making it any shorter, only longer. But I'm not sure any of those three options really describe what it does, nor I can think of a better option ... `processZeroDeltaCycle()`? Sounds horrible.

I kind of like it the way it is though. We're processing a cycle and our `dt` is 0: `process([dt=]0)`. 

There's nothing special about it, the delta just happens to be 0. I don't think I'd even make the delta optional / default to 0.

test/test_SimulatedLinkamT95.py
MikeHart85: It implicitly asserts no exceptions are thrown when calling `pumpCommand` with valid parameters.

I would put a `try: ... except: self.fail(...)` around it, like the first three tests... would that be better?

MichaelWedel: Either that (should probably go into a function `assertRaisesNothing` or similar) or actually checking the status each time. Or is the pump speed not part of the status bytes?

On the other hand if the method can raise an exception there should probably be a test that actually causes this exception and uses `assertRaises` to detect this.

test/test_SimulatedLinkamT95.py
MichaelWedel: I'm not sure it isn't special, because in 23 of 36 cases in this file the argument is 0. Maybe thinking about the reason for doing this leads to a good name.

MikeHart85: Hm... Basically, I needed a cycle in order to process transitions, but I didn't want any time to pass.

But the only reason I didn't want any time to pass is to make it easier to synchronize when to assert the target temperature has been reached. 

I could have just as well used `0.1` (or any other value shorter than the 60 second target runtime)... but then I would have had to do more math to calculate how long I need to process to reach (or not reach, or overshoot, etc) the target time/temperature, especially when multiple transitions needed to occur before I got to that point.

In some places, it doesn't even matter for timing. For example, the `process` calls immediately after creating the `linkam` instance and the ones following the `getStatus` immediately after instantiation. Those `process` calls I could have written `process(0x1337BEEF)` and it wouldn't have made any functional difference.

test/test_SimulatedLinkamT95.py
MikeHart85: Pump speed is part of the status bytes (though not whether we're in auto or manual mode; there is no way to check that), but the way that speed changes is likely wrong in the current implementation, so I'm not sure it's worth testing for.

Basically, the pump command is only partially implemented and there is currently little to no interest in expanding on it since it is practically never used in manual mode, only automatic.

Nevertheless, there is some code, and so I just called the command with all valid values to make sure it doesn't crash (and to satisfy the coverage system).

The pump command does not explicitly raise any exceptions, nor would I expect it to under normal circumstances. The test is to ensure there are no unexpected exceptions.

test/test_SimulatedLinkamT95.py
MichaelWedel: So the reason is to process transitions without passing time. The "any time is valid" is another pointer in my opinion. If time doesn't matter in a time based simulation then it probably has nothing to do with time and it shouldn't be necessary to supply a time to do what you want to do.

Do I understand correctly that it is basically "skipping to the next point where time matters again"?

test/test_SimulatedLinkamT95.py
MichaelWedel: Okay, I'd say in that case it's best to add the `assertRaisesNothing`-thing to make clear that **that** is tested. What about checking that invalid speeds are ignored (if that is the device behavior)?

test/test_SimulatedLinkamT95.py
MikeHart85: No, that's not correct.

The only thing that all `process(0)` calls have in common is that I want to process a cycle (specifically to trigger any transitions that may be valid, but that applies to all `process` calls regardless what the delta t parameter is) with a delta t of 0.

In some instances where `process(0)` is called, time does not matter at all, because it just so happens that the state the statemachine is in doesn't have any kind of time-based behaviour.

In some instances where `process(0)` is called, time does not matter for the purposes of the particular test I am running, because even though the state has time-based behaviour, it is neither behaviour I am testing nor does it have side-effects on behaviour I will subsequently test.

In some instances where `process(0)` is called, time matters and I want the delta to be 0 to make the delta in a subsequent `process` call both easier to calculate and clearer in intention for the reader. For example:

``` python
# Cool for almost a minute but not quite
linkam.process(59.5)
```

I could have used `process(30)` instead of `process(0)` in the previous three `process` calls. But then I would have had to write `process(29.5)` here. 

The reason that **three** previous `process(30)` calls add up to a difference of -30 in the fourth is because the first two don't matter and the third one does. 

The reason the first two don't matter is because I hadn't started cooling yet. They occurred in the different states, the behaviour of which I wasn't testing. The third one matters because I already started cooling and cooling is the state I'm testing here.

By using 0s in all three cases, I keep things simple and can focus on the forth -- the thing I'm actually testing.

test/test_SimulatedLinkamT95.py
MikeHart85: Where would you add that?

Arguably, a function like that should be provided by `unittest.TestCase` (and it is a fairly common complaint that such a thing doesn't exist).

It doesn't really make sense to make a helper function like that in `TestSimulatedLinkamT95`, since chances are we'll want to use it in other tests as well.

A free-standing function in `core.utils` maybe? We'll still have to pass in the TestCase instance in order to call `TestCase.fail` when an exception is caught, right? Should we also pass in an expected exception type, the same way `TestCase.assertRaises` does?

Edit:

Yes, as far as I know, device behaviour is to just ignore invalid speed (same as any invalid commands or parameters).

test/test_SimulatedLinkamT95.py
MichaelWedel: > The only thing that all process(0) calls have in common is that I want to process a cycle

Maybe something like `nextCycle` then. The importance of this is nowhere near the length of this discussion, I just thought it could be nice to make the intent of this more clear. After all `process(0)` could be implemented such that it is a no-op that doesn't result in going to the next cycle.

test/test_SimulatedLinkamT95.py
MichaelWedel: Since it's only for testing I'd put it in `test/__init__.py` or create a `test/utils.py`, what do you think? I like the idea of passing `self` to that function to call `TestCase.fail`. I don't think passing an exception type is necessary, I don't expect any exception, so any exception raised violates that expectation.

MichaelWedel: I think there is something in Mantid that does that (from `Framework/PythonInterface/test/__init__.py`):

```
# Case difference is to be consistent with the unittest module
def assertRaisesNothing(testobj, callable, *args, **kwargs):
    """
        unittest does not have an assertRaisesNothing. This
        provides that functionality

        Parameters:
            testobj  - A unittest object
            callable - A callable object
            *args    - Positional arguments passed to the callable as they are
            **kwargs - Keyword arguments, passed on as they are
    """
    try:
         return callable(*args, **kwargs)
    except Exception as exc:
        testobj.fail("Assertion error. An exception was caught where none was expected in %s. Message: %s"
                     % (callable.__name__, str(exc)))
```

test/test_SimulatedLinkamT95.py
MikeHart85: > After all `process(0)` could be implemented such that it is a no-op that doesn't result in going to the next cycle.

Hm... I would find this very surprising (to the point where I would raise an issue against it).

If I tell a system to process a cycle, I expect it to process a cycle (or throw an exception / return an error code). I suppose this comes back to something you said earlier:

> If time doesn't matter in a time based simulation then [...]

It's not really a time based simulation. It's a cycle based simulation. Time is just a parameter that gets passed along "to whomever it may concern". When I call `process(0)`, I'm saying: "Process a cycle. Oh, and by the way, if anyone asks, tell them dt is 0."

If that's the part that is confusing, maybe `process` itself isn't a good name?

But, as you say, it's questionable whether this is worth much effort. You would only call `process` manually in tests, anyway. It's called automatically during normal simulation. And while, normally, `dt` wouldn't be 0 since it loops and calculates a delta, there is nothing special about a delta of 0 either. The `dt` argument is -- and should only ever be -- used as a factor. Never as a direct condition (though things increased as a result of factoring in `dt`s may of course be conditions).

test/test_SimulatedLinkamT95.py
MikeHart85: Nice find! I had no idea that was there.

You're right, exception type shouldn't matter.

I'll cheekily steal that function from Mantid and put it in `plankton/test/__init__.py`.

test/test_SimulatedLinkamT95.py
MikeHart85: Thinking about this some more, I don't think this is that trivial. It bothers me that it bothered you, or that you even noticed it to begin with.

A `process(0)` call shouldn't have stood out at all. I would have thought anyone reading that would immediately see that it's doing a cycle with a delta of 0, and that there's nothing peculiar about that. A big part of the reason I don't really want a separate function for it is **because** there's nothing special about it and I don't want to send a message that there is.

If you thought it was weird or some sort of special case, others likely will as well. And we do expect (at least some) users will write unit tests for their devices, so they will (or should) be calling / seeing it too.

MichaelWedel: > It's not really a time based simulation. It's a cycle based simulation.

I fell into the trap :) You're totally right, it's cycle-based not time based. I got myself quite confused apparently. After that reminder I actually agree that there shouldn't be a separate function for this, not at all, and yes, maybe we can find a better name instead of `process`.

You're also right about `process` never being called anywhere except in unit tests, but there it might be called with "I don't care" often (as seen here). So what about making `dt` optional, possibly with `0` as default?

I think even though this was off-topic for the PR it was good that we cleared that up somehow, if anything I re-learned that it's based on cycles and not time :D

test/test_SimulatedLinkamT95.py
MichaelWedel: I remembered that I used it before...I think it's okay to steal it but it should maybe be documented where it was stolen from :)

PR #85, opened 2016-08-29
Strip newlines from StreamHandler

OwenArnold:
@MikeHart85 stripping the newlines from the input stream makes life a lot easier when connecting via a telnet client.

### DIFF
--- a/adapters/stream.py
+++ b/adapters/stream.py
@@ -41,7 +41,7 @@ class StreamHandler(asynchat.async_chat):
         self.buffer.append(data)
 
     def found_terminator(self):
-        request = ''.join(self.buffer)
+        request = ''.join(self.buffer).strip()
         reply = None
         self.buffer = []
 


OwenArnold:
I think that this is supposed to work via `in_terminator`, `out_terminator` specifications in the bindings, but even with the `toggle crlf` modes applied. I couldn't reach the right combination that stripped out the `\n` from my commands.

MichaelWedel:
I'm not sure this is the right thing to do. Telnet seems to specify CRLF as EOL-character, but the device is not talking telnet. Telnet is just very similar by chance. I made this small client: https://gist.github.com/MichaelWedel/fd288ac19ef1d08d994e639be78aadd0

Connect to the device like this:

```
$> python stream_client.py -x -c localhost 9999
```

MichaelWedel:
The original gist had some very unfortunately named variables, I made an update.

MikeHart85:
I agree with @MichaelWedel, this doesn't feel like the right way to deal with this. Sending the wrong line-endings is technically an error... If we sanitize the input, we're potentially hiding an issue.

Allegedly this is supposed to force telnet to use `\r` (CR) as a line terminator (relying on the fact that if you copy-paste text with CR, telnet will send it exactly as-is without adding NL):

```
$ sed -u 's/$/\r/' | telnet [IP] [PORT]
```

But I couldn't get it (or any variation of this strategy using `tr` or `paste`, or replacing `telnet` with `nc`) to work for me on Ubuntu. Maybe it will give one of you guys ideas though.

OwenArnold:
@MichaelWedel @MikeHart85. Having read your responses, I am forced to agree that this (the changes I provided) is not the right fix. 

The stream editor approach above didn't work for me. I haven't tried @MichaelWedel's tool yet, but I do like the idea of interacting with the device + adapter using the native interface via some kind client. I'm sure that this is beneficial for development and deployment. Should we think about distributing these kinds of tools alongside plankton?

MikeHart85:
I like the idea of shipping a tool like this with plankton. It could be in the same image, and even part of plankton itself. That way, you could just specify the device you're connecting to and it could figure out the correct line-endings to use by itself.

MichaelWedel:
What's a good way to handle a protocol like the Linkam with all this status byte stuff? In the example above this just gets printed as `0x80` and so on in lack of a better solution.

OwenArnold:
@MichaelWedel @MikeHart85 I think this issue goes away with the work that @MichaelWedel will be doing in the `control.py` RPC direction. I'm closing this PR.

PR #86, opened 2016-08-31
Chopper should check states before triggering command switches

### LINKED ISSUE #84: Fix behavior of command switches
After checking back with Andres from the chopper group, we have to revisit how the command switches work. While the chopper **controller** showed the current behavior for at least some of the commands, Andres implemented checks in the **CHIC** that prevent command switches to be triggered in states where it's not allowed. This means that we have to use the state machine for these checks and only allow commands in the proper states, since were simulating at the CHIC-level.

MichaelWedel:
This fixes #84.

To test you could start the chopper (in the sense of setting a speed and a phase) and then issue a `deinit`-command. Like before, nothing happens. If you then issue `stop`, it should not go directly to the `init`-state as it did before.

This device needs unit tests at some point.


test/test_StateMachine.py
```diff
@@ -367,3 +367,34 @@ def test_reset(self):
...
+
+        sm.process(0)
+        self.assertEqual(sm.state, 'foo')
+        self.assertEqual(sm.can('bar'), True)
+        self.assertEqual(sm.can('init'), False)
+
+        sm.process(0)
+        self.assertEqual(sm.state, 'bar')
+        self.assertEqual(sm.can('foo'), True)
+        self.assertEqual(sm.can('init'), True)
```
MikeHart85: Arguably, these could use `self.assertTrue()` and `self.assertFalse()` respectively. Another alternative would be `self.assertIs()`.

I don't really have a preference one way or the other, just something to consider.

`self.assertIs()` would have the benefit of testing for identity with `True` and `False`; all the other asserts will succeed even if the value is only "truthy" or "falsy" (1 == True, 0 == False, etc).

MichaelWedel: That's a good point, thanks. I used `assertEqual` so that you can see what it returns. It sounds like `assertIs` is the thing that really should be used here.

test/test_StateMachine.py
```diff
@@ -367,3 +367,34 @@ def test_reset(self):
...
+
+        self.assertEqual(sm.state, None)
+        self.assertIs(sm.can('init'), True)
+        self.assertIs(sm.can('foo'), False)
+        self.assertIs(sm.can('bar'), False)
+
+        sm.process(0)
+        self.assertEqual(sm.state, 'init')
+        self.assertIs(sm.can('foo'), True)
+        self.assertIs(sm.can('bar'), False)
```
MikeHart85, line 392: What about the case when you're in `'init'` and you call `can('init')` (and the same for other states)?

That should always return `False`, right? Theoretically, I think the StateMachine does support a transition that loops back on itself... though I can't really think of a scenario where it would be useful. Every cycle calls `in_state` anyway, but that would force an `on_exit` and `on_entry` when the transition is triggered.

Anyway, do you think it's worthwhile testing for `self.assertIs(sm.can(sm.state), False)` here?

MichaelWedel: Ah you're right, I forgot to test these. But since it's in principle possible to have 'self-transitions' and `on_exit` and `on_entry` may perform meaningful work even for such a transition I think I'd prefer adding this scenario instead of "forbidding it".

MikeHart85: I think you're right, no reason to forbid it... never know if it may become useful.

Looks good!

PR #88, opened 2016-09-08
Powersupply and modified chopper for testing purposes

MikeHart85:
This is a branch a made quite a while back for demoing at the workshop.

It contains a simple power supply device (roughly put together based on a quick conversation with John Holt) and some tweaks to the chopper to add randomization to the phase.

It's quite out of date now and cannot be auto-merged (if we would even want to, since this is not a real device).

We should decide what we want to do with this, if anything. Perhaps it can be used as a starting point for a real power supply?

### DIFF
--- a/devices/chopper/defaults.py
+++ b/devices/chopper/defaults.py
@@ -18,6 +18,7 @@
 # *********************************************************************
 
 from core import State
+import random
 
 
 class DefaultInitState(State):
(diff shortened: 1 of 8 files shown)


MikeHart85:
Closing this as it is very stale and devices have changed / been simplified significantly for the 1.0 release.

PR #90, opened 2016-09-12
Separate remote control of simulated device

### LINKED ISSUE #39: Runtime control for simulation
As per the discussion in #35, we need to think about how to implement runtime control of the simulation.

My current suggestion would be to pull the `Controllable` and `ControlServer`-concepts out of the live data prototype, because these two solved the exact same problem in that context.

From my point of view separating the control channel from the simulated device channel is a good thing, because that means implementing the control capability once, instead of once for each device protocol. This does however require that we refine the `Adapter` base class so that a potential server can query the adapter about its capabilities and expose the appropriate interface. Again this overlaps to some degree with the live data prototype where the exposed objects are quite diverse, but it is possible to query them easily.

MichaelWedel:
This fixes #39.

The simulated device can now be accessed through JSON-RPC 2.0 via ZMQ if enabled by the new `-r` in the startup script. Exposure happens on the object level, so classes don't have to be modified in any way to use this functionality.

On the client side, each exposed object is wrapped in a proxy that transforms access to properties, methods and plain attributes into JSON-RPC-requests. The proxy is more or less transparent and exceptions on the server side are re-raised on the client side if they are in `exceptions` (Python 2) or `builtins` (Python 3).

A new command line utility has been added, called `control.py`.

The new functionality is described in the readme.


MichaelWedel:
Since opening the pull request not much has changed, I just took up results from coveralls, added more tests and tweaked a few method parameters. From my point of view this is okay now.

requirements.txt
```diff
@@ -1,2 +1,4 @@
 pcaspy
 six
+zmq
```
MikeHart85: Wouldn't it be better if this was `pyzmq`?

`zmq` appears to be an alias to it, but always reports version "0.0.0".

(same in requirements-dev.txt)

core/rpc_client.py
```diff
@@ -0,0 +1,230 @@
...
+        if 'error' in response:
+            if 'data' in response['error']:
+                exception_type = response['error']['data']['type']
+                exception_message = response['error']['data']['message']
+
+                if not hasattr(exceptions, exception_type):
+                    raise JSONRPCServerSideException(exception_type, exception_message)
+                else:
+                    exception = getattr(exceptions, exception_type)
+                    raise exception(exception_message)
```
MikeHart85: Could `exceptions` here be replaced by `__builtins__`? 

That would also allow us to get rid of the ugly import issues with the exceptions module and Python 2 vs 3.

Edit: Trying this out, the unit test for it fails because the general exception is raised instead of an `AttributeError`... but I'm not sure why. In a Python console (both 2.7 and 3.5), calling `hasattr(__builtins__, 'AttributeError')` returns `True`. So it should work...?

MichaelWedel: I think I also tried that along the way, but it didn't work. In Python 3 the module is actually called `builtins`, not `__builtins__`. In the end I settled for the "ugly import".

MikeHart85: This works (replacing the `if not hasattr` block, no `import`s required at all):

``` python
if exception_type in __builtins__:
    exception = __builtins__[exception_type]
    raise exception(exception_message)
else:
    raise JSONRPCServerSideException(exception_type, exception_message)
```

However, this relies on, and goes against advice of, this Python feature: https://docs.python.org/3/reference/executionmodel.html#builtins-and-restricted-execution

Ironically, in spite of the advice against using it, `__builtins__` is by far the most reliable and universally supported way to do what you want here.

Not sure what the lesser evil here is.

MichaelWedel: Thanks for finding that solution. Does it work for 2 and 3? I looked at the `__builtins__` warning in the link you provided, they advise to use the `import builtins` route. I'm leaning towards leaving it like it is, since we have other places where we do something like this for the sake of supporting Python 2 and 3.

I'm wondering if there's a more elegant solution altogether, but since whatever we do here will end up having the same functionality I don't think it's worth spending large amounts of time on.

MikeHart85: Yes, `__builtins__` is always available (without importing anything) and works exactly the same in both Python 2 and 3.

The one catch with it is that it works differently in the main (executed) module than in all other (imported) modules. In the main module it is an alias for the `__builtin__` module (in Python 2) or the `builtins` module (in Python 3). In all other modules, it is an alias for `__builtin__.__dict__` or `builtins.__dict__`, respectively. Odd behaviour, but consistent across all Python versions.

I'm happy leaving it as is... the only reason it caught my eye was that PyCharm flags `import builtins as exceptions` as an error.

MichaelWedel: That's interesting, mine did not complain at all, neither on `exceptions` nor on `builtins as exceptions`. Maybe it's because I have Python 3 set as the interpreter for the project. It does flag an error on importing `mock`.

Well anyway, if you're happy to leave it I suggest we keep it in mind and if we stumble across a better solution we can still do that later.

MikeHart85: I think it comes up as an error on mine because I enabled Python 2 vs 3 compatibility checks in Settings -> Inspections (and it doesn't realize the try-except is there to deal with just that).

But yeah, it's fine as is... just wanted to see if there's a different way and that's the only real alternative I could find.

core/rpc_client.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        An exception type for exceptions related to the transport protocol, i.e.
+        malformed requests etc.
+
+        :param message: A message describing the exception.
+        """
+        self.message = message
+
+    def __str__(self):
+        return self.message
```
MikeHart85: Wouldn't this work exactly the same if you just did:

``` python
class JSONRPCProtocolException(Exception):
    """docstring"""
    pass
```

`Exception` already does these things for you.

Edit: Actually, you can even leave off the `pass` if you have a doctring like that.

core/rpc_client.py
```diff
@@ -0,0 +1,230 @@
...
+        the message are stored.
+
+        :param type: Type of the exception on the server side.
+        :param message: Exception message.
+        """
+        self.type = type
+        self.message = message
+
+    def __str__(self):
+        return 'Exception on server side of type \'{}\': \'{}\''.format(self.type, self.message)
```
MikeHart85: Similar to the comment on the other exception, you could do just a `pass` (and expect the user to format their own message). Alternatively, if you want to keep the usage as-is:

``` python
class JSONRPCServerSideException(Exception):
    def __init__(self, type, message):
        """docstring"""
        super(JSONRPCServerSideException, self).__init__('Exception on server side of type \'{}\': \'{}\''.format(type, message))
```

Optionally, could also keep the `self.type` and `self.message` assignment after the `super` if you want them accessible.

core/rpc_client.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class ZMQJSONRPCConnection(object):
+    def __init__(self, host='127.0.0.1', port='10000'):
+        self._socket = self._get_zmq_req_socket()
+        self._socket.connect('tcp://{0}:{1}'.format(host, port))
+
+    def _get_zmq_req_socket(self):
+        context = zmq.Context()
+        return context.socket(zmq.REQ)
```
MikeHart85: Couldn't this just be in `__init__`?

It seems to only be used here, and amounts to at most one more line in `__init__`.

core/rpc_client.py
MichaelWedel: Thanks for pointing that out, it's a bit inconsistent. I think the original type and message should be kept, but message should be what `__str__` returns. So the member can't be `message`. I'll go and update that.

core/rpc_client.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def json_rpc(self, method, *args):
+        """
+        This method takes a ZMQ REQ-socket and submits a JSON-object containing
+        the RPC (JSON-RPC 2.0 format) to the supplied method with the supplied arguments.
+        Then it waits for a reply from the server and blocks until it has received
+        a JSON-response. The method returns the response and the id it used to tag
+        the original request, which is a random UUID (uuid.uuid4).
+
+        :param socket: ZMQ REQ-socket.
```
MikeHart85: No parameter called socket in this method (anymore?).

control.py
```diff
@@ -0,0 +1,86 @@
...
+    else:
+        if not arguments:
+            return attr
+        else:
+            setattr(remote[object_name], method, *args)
+
+
+parser = argparse.ArgumentParser(
+    description='A client to manipulate the simulated device remotely through a separate channel. The simulation must be started with the --rpc-host option.')
+parser.add_argument('-i', '--ip', default='127.0.0.1', help='IP-adress of the host plankton is running on.')
```
MikeHart85: Minor typo: IP a**d**dress

core/rpc_client.py
MichaelWedel: It used to be in init, but testing got easier with this additional method, for tests where it's needed, this method is patched to return a `Mock`.

core/rpc_client.py
MichaelWedel: Yes...I wrote the docstrings too early I suppose and then didn't check properly. Thanks.

control.py
MichaelWedel: Oooops, thanks :) I'm blaming my native language processing firmware where one `d` is perfectly acceptable ;)

core/rpc_client.py
```diff
@@ -0,0 +1,224 @@
+#!/usr/bin/env python
```
MikeHart85: Shouldn't need this unless this file is meant to be executed directly from the command line.

`rpc_server.py` and both unit test files also shouldn't need this, I think.

Only `control.py` does need it, otherwise strange things happen when you try to run it directly.

control.py
MikeHart85: Always the firmware's fault... ;)

core/rpc_client.py
MichaelWedel: Ah this must be a leftover from the olden days where these files lived directly in the top level directory and were scripts to execute directly. Or it got in with the license headers.

core/rpc_client.py
MikeHart85: Uff... that makes me a bit uneasy. Understandable, of course... just the thought of pulling apart code that belong together for the sake of tests... feels dirty.

Not a big deal here though.

MichaelWedel: It's the usual problem of testing things that involve networking, file access and so on...at least in my experience that is usually hard and/or inconvenient to do. In principle `zmq.Context` could be mocked here instead, I suppose the outcome would be similar.

core/rpc_client.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        This class serves as a base class for dynamically created classes on the
+        client side that represent server-side objects. Upon initialization,
+        this class takes the supplied methods and installs appropriate proxy methods
+        or properties into the object and class respectively. Because of that
+        class manipulation, this class must never be used directly.
+        Instead, it should be used as a base-class for dynamically created types
+        that mirror types on the server, like this:
+
+            proxy = type('SomeClassName', (ZMQJSONRPCObjectProxy, ), {})(socket, methods, prefix)
```
MikeHart85: The `socket` here is a bit confusing... shouldn't it be `connection`?

README.md
```diff
@@ -193,3 +193,61 @@ $ telnet 192.168.99.100 1234
 
 This `-p` argument links port 4321 on the container to port 1234 on the VM network adapter. It must appear after `docker run` and before `dmscid/plankton`. This allows us to connect to the container from outside of the VM, in this case using Telnet. The `192.168.99.100` IP is the IP of the VM on the bridge network between the host and the VM. VirtualBox will typically use this IP when available, but it may be different on your system.
 
+## Remote access to devices
+
+*Please note that this functionality should only be used on a trusted network.*
```
MikeHart85, line 198: This is because the traffic is not encrypted? Would it be possible/difficult to encrypt it? If it's just a flag or something in `zmq`, it might be worthwhile since the cost would be negligible.

README.md
```diff
@@ -193,3 +193,61 @@ $ telnet 192.168.99.100 1234
...
+Besides the device specific protocols, the device can be made accessible from the outside via JSON-RPC over ZMQ. This can be achieved by passing the `-r` option with a `host:port` string to the simulation:
+
+```
+$ python simulation.py -r 127.0.0.1:10000 -d chopper -- -p SIM:
+```
+
+Now the device can be controlled via the `control.py`-script in a different terminal window. The service can be queried to show the available objects using the `-l` or `--list-objects` flag:
+
+```
+$ python control.py --ip 127.0.0.1 --port 10000 --list-objects
```
MikeHart85: Wouldn't it be better if the IP/Port format matched the server side?

So if we replaced `--ip` and `--port` with `-r`/`--remote` like so:

```
$ ./control.py -r 127.0.0.1:10000 --list-objects
```

README.md
MichaelWedel: No, I was more concerned with clients having direct access to the object on the server with no authentication. In fact this warning might apply to all our network functionality, but I hadn't thought about it so far, because the normal use case is that these things run in a container or local net anyway.

README.md
MichaelWedel: Maybe you're right, although I would have liked to us I/p on the server side, but p is occupied. So for being consistent we should use r on the client.

README.md
```diff
@@ -193,3 +193,61 @@ $ telnet 192.168.99.100 1234
...
+```
+$ python control.py device targetSpeed 100
+$ python control.py device start
+```
+
+Only numeric types and strings can be used as arguments via the `control.py`-script. The script always tries to convert parameters to `int` first, then to `float` and leaves it as `str` if both fail. For other types and more control over types, it's advised to write a Python script instead using the tools provided in `core.rpc_client` which makes it possible to use the remote objects more or less transparently. An example to control the chopper:
+
+```
+from time import sleep
+from core.rpc_client import ZMQJSONRPCConnection, get_remote_object_collection
```
MikeHart85: I don't think I'm a fan of the class names.

As far as I can tell, from the user's perspective, ZMQ and JSONRPC are implementation details that are never seen or interacted with... except for that a lot of things have it in the name. Additionally, since every module is a namespace in Python, it seems a bit redundant to have such long prefixes. And `ExposedObject` and `ExposedObjectCollection` don't follow that format anyway (I rather like the way those are named).

How would you feel about renaming things to, for example:

```
JSONRPCServerSideException  ->  RemoteException
JSONRPCProtocolException    ->  ProtocolException
ZMQJSONRPCConnection        ->  ControlClient
JSONRPCObjectProxy          ->  ObjectProxy
ZMQJSONRPCServer            ->  ControlServer
```

And maybe also 

```
rpc_client.py    ->    control_client.py
rpc_server.py    ->    control_server.py
```

Additionally, I would make `get_remote_object_collection` and `get_remote_object` members of `ZMQJSONRPCConnection` (or rather `ControlClient`). 

That way, user facing code becomes:

``` python
from plankton.core.control_client import ControlClient

remote = ControlClient(host='...', port='...')
chopper = remote.get_object('chopper')

# or maybe even just...
chopper = remote['chopper']
```

This feels a lot more intuitive / readable to me.

control.py
```diff
@@ -0,0 +1,86 @@
...
+    description='A client to manipulate the simulated device remotely through a separate channel. The simulation must be started with the --rpc-host option.')
+parser.add_argument('-i', '--ip', default='127.0.0.1', help='IP-address of the host plankton is running on.')
+parser.add_argument('-p', '--port', default=10000, help='Port on which the RPC-server is listening.')
+parser.add_argument('-l', '--list-objects', action='store_true',
+                    help='List all objects exposed by the server and exit.')
+parser.add_argument('-a', '--show-api', action='store_true',
+                    help='List all properties and methods of the controlled object.')
+parser.add_argument('object', nargs='?', default='device', help='Object to control.')
+parser.add_argument('member', nargs='?', help='Object-member to access.')
+parser.add_argument('argument', nargs='*',
```
MikeHart85: Nitpick: should be `arguments` (or maybe `params` to avoid the `args.arguments` awkwardness)

README.md
MichaelWedel: I agree on the naming. `ExposedObject` was named strangely before as well (and stuff was differently organized as well). In fact I already renamed everything according to your suggestions, I like those names. They fit better 

I also agree that those methods should be moved into `ControlClient`. The problem with the `dict`-like interface is that the server does not necessarily expose a collection of named objects, it can also expose a single object, which is then unnamed. In that case (and I think it's a valid use case) it would not make sense to access it like a `dict`, this is basically the only reason why those two methods exist separately. What I do not like about this is that you have to know it beforehand, at least partly. Since `ExposedObjectCollection` inherits from `ExposedObject`, you can always go that route (using the current scheme):

``` python
obj = get_remote_object(connection)
object_names = obj.get_objects() # May fail if obj is not in fact a collection

objects = [get_remote_object(connection, name) for name in object_names]
```

So the client needs to do some sort of check what the server actually exposes, so it might not be as easy as a `dict`-like interface.

My suggestion would be to move `get_remote_object` into `ControlClient` and keep a small function that creates a dict from that object, like `get_remote_object_collection` does now:

``` python
connection = ControlClient(host='...', port='...')
objects = remote_collection(connection.get_object())
```

The other alternative I see would be that the `dict` like interface is just empty if the remote object is not of the right type, but at the moment that does not convince me, because you still need a way to get the top level object, so there are two interfaces in one...

MichaelWedel: I'm just experimenting a bit with the `dict`-like thing...it's not particularly pretty, because for the top level object you would have to do `remote['']`, which seems a bit unexpected. I'll try a few other things, but in the end it might be the least strange thing to keep a helper function that constructs an object `dict` (as it is now).

MichaelWedel:
Okay, this is what I propose for now:

``` python
from control_client import ControlClient, remote_objects

client = ControlClient(host='...', port='...')
objects = remote_objects(client)
chopper = objects['chopper']
```

For scripts that run once and do not wish to re-use the `ControlClient`-instance (like `control.py`) it can be shortened to:

``` python
objects = remote_objects(ControlClient(host='...', port='...'))
chopper = objects['chopper']
```

Objects can still be pulled out one by one from the `ControlClient`:

``` python
client = ControlClient(host='...', port='...')

# top level object, this should always work
top_level_object = client.get_object()

# object further down the hierarchy, works only if server side actually exposes that
chopper = client.get_object('chopper')
```

In principle the `remote_objects` is not needed, but it is convenient to retrieve all remote objects at once, so I'd like to keep it.

MichaelWedel:
I'm going to update the readme once we agree on whether the client side is okay.

README.md
MikeHart85: Hm... I might be misunderstanding something about how the single-object scenario works. But just looking at the way it is now:

``` python
class ControlClient(object):
    def get_object(self, object_name='')
        pass

def remote_objects(source, object_name=''):
    """
    :param source: A source of remote objects, must be of type ControlClient.
    """
    if not isinstance(source, ControlClient):
        raise RuntimeError('A ControlClient instance is required to retrieve remote objects from the server side.')
```

The first argument to `remote_objects` is already necessarily a ControlClient. So why can't that be turned into a member method, with the first argument being `self`? Don't need the `isinstance` check that way, and it feels conceptually correct from an object oriented perspective (you want the remote objects from that client... `ControlClient.remote_objects()`).

For the `dict` notation, I would see that as just a convenience alias for `get_object`. So I don't see the harm in adding it (and keeping the explicit `get_object`). That way you can use whichever is more readable for what you want to do:

``` python
class ControlClient(object):
    def get_object(self, object_name='')
        pass

    def get_objects(self, object_name=''):
        object_names = self.get_object(object_name).get_objects()
        return {obj: self.get_object(obj) for obj in object_names}

    def __getitem__(self, object_name):
        return self.get_object(object_name)
```

> The other alternative I see would be that the dict like interface is just empty if the remote object is not of the right type, but at the moment that does not convince me, because you still need a way to get the top level object, so there are two interfaces in one...

You're right, of course... it's a bit awkward that sometimes the dictionary notation just doesn't make sense, depending on what the server is hosting. But isn't _that_ the source of the issue? That the server is exposing things in an inconsistent way?

Can't we just say the server always hosts a collection at the top level. And if you just want to expose one object, then you just expose one object... but it still has to have a name and be inside the top level collection. What would be the downside of that?

Edit: Renamed `remote_objects` to `get_objects` when as a member. Could also be `get_object_collection` if we want it to be more visually distinct.

core/control_client.py
```diff
@@ -0,0 +1,206 @@
...
+        self.server_side_type = type
+        self.server_side_message = message
+
+
+class ProtocolException(Exception):
+    """
+    An exception type for exceptions related to the transport protocol, i.e.
+    malformed requests etc.
+    """
+    pass
```
MikeHart85: Nitpick / my fault: This `pass` could be removed (docstring already satisfies need for a statement)

test/test_control_client.py
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env python
```
MikeHart85: These tests shouldn't need the shebang either (they're never executed directly, right?)

Same for the test_control_server.py file.

MichaelWedel: True. I'm almost certain this came in as a copy paste error when I inserted the license headers.

README.md
MikeHart85: Just to provide some usage samples as I imagine them (assuming we keep the feature where the server exposes a single unnamed object):

``` python
from control_client import ControlClient

remote = ControlClient(host='...', port='...')

# If I know only a single unnamed object is exposed
chopper = remote.get_object()

# If I know a collection that contains a 'chopper' is exposed
chopper = remote['chopper']

# ... or I can spell it out if I really want to
chopper = remote.get_object('chopper')

# If I'm not sure what's there and I want to browse using an actual `dict`
# (does this throw an exception if only a single object is exposed?)
objects = remote.get_objects()
```

MichaelWedel: The server always exposes one object at the top level (so in this it is consistent), which happens to be an `ExposedObjectCollection` in case of `simulation.py`. This object has the capability of dispatching the JSON-RPCs down to the objects it is holding. In fact you can expose collections of collections of collections and on the client that would translate to `client.get_object('object.object.object')`.

I don't see a reason to enforce exposing a collection at the top level on the server side and relying on that on the client side, I'd rather resolve that on the client side in some other way.

What about having a different implementation of `ControlClient` that behaves in the `dict`-way?

MichaelWedel: Ah I hadn't seen your usage example. I thought about that solution (`get_object_collection` raises an exception if not applicable), I suppose it's more or less equivalent to using an external function, which would also behave that way.

So let's say we agree on that solution?

MichaelWedel: Okay, now I realized the `dict`-interface is back :D 

How about just this:

``` python
client = ControlClient(host='...', port='...')

root_object = client.root()
chopper = client['chopper']
```

Maybe that name is better than `get_object`? Do you maybe have another suggestion?

The `dict`-part of the interface can be implemented using the assumption of `get_objects` on the server side, but it will raise exceptions if that's not the case. Then the explicit `dict` is not required at all, which makes the interface simpler.

core/control_client.py
MichaelWedel: For these one-line changes which are obvious and don't need discussion, please feel free to just modify directly on the branch (I often do that via github directly).

README.md
MichaelWedel: I'm just implementing the `dict`-part and I think it doesn't really fit, because of the possibility of having nested collections. You could do `client['object.object']` which looks strange.

I now think the best compromise between having a nice interface and spending too much time on this is to provide `get_object_collection` as member on `client` which may fail and `get_object` which always works if used without parameters.

MichaelWedel:
Let me know if you can live with this solution for the time being. I don't really see the dict-interface working very well at the moment, maybe that can be solved at some later point. Given that this solution does what we need at the moment I'd rather continue with the other aspects of the re-design and draw this to an end.

MikeHart85:
Looks perfectly fine to me. You're right, should move forward with some more bigger changes... we can work out details later if it becomes necessary.

PR #93, opened 2016-09-16
Remove run-logic from Adapter

### LINKED ISSUE #35: Modify startup script to specify delta t
The startup script for simulations needs to accept delta t as a parameter from the user so that simulations can be run at different speeds.

This is important later on when we want to run different scenarios as a kind of "system test suite".

### LINKED ISSUE #87: Ibex back-door requests
We met with the IBEX team 5th-September 2016 to discuss their requirements for Plankton.

Use-cases to support:
1. "How long has this taken". Just to track the time. Time since the emulation started. Doesn't mean that a speed up in needed. Simulation time should be reported not real time.
2. "How long will this take" . Estimating the total execution time a script will take - Possibly the same as above but with a speed up involved. Simulation time should be reported not real time.
3. Having a robust way to incrementally develop an IOC - Therefore being able to run an manipulate a device emulator for the purposes of manual development testing. Might involve disabling the state machine.
4. "Unit testing" production IOCs - Therefore using device emulators via a back-door. IOCs take 10s of seconds to start up and stop, so they can only be initialised once per test-suite. Might involve switching a setup at runtime via the back-door.

Notes:
- Start & Stop scripts rather than a dedicated service.
- Plankton will need to "Disconnect" and "Reconnect" via the back-door.
- Definition of back-door: Ability to control general environment parameters as well as specialised device and it's attributes.
- Plankton must give access to the `dt factor` and `cycle rate`.

### LINKED ISSUE #92: Move run-logic to new class Environment
In the current design it's the Adapter's responsibility to run whatever service is required to expose the device in a sort of "central processing loop". Instead, a new class `Environment` will be created which handles this main loop and asks the Adapters to process.

MichaelWedel:
This fixes #92, fixes #35, fixes #87.

It is now no longer the responsibility of the adapter to actually start and run the simulation, this has been moved into `SimulationEnvironment`, as discussed in the design document earlier.

`SimulationEnvironment` provides control over the processing time in each cycle and how that maps to simulated time via a factor. It also keeps track of the total processing time, total simulation time and number of processing cycles as well as simulation cycles. It is possible to pause and resume the simulation. The simulation can also be stopped, causing the program to exit (not entirely sure this is exactly what we want).

All these things are exposed via the control server (the object is named "simulation" and is separate from "device") so that manipulation at run time is possible.

Documentation has not been added yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,4 +11,3 @@ RUN pip install -r plankton/requirements.txt && \
 COPY . /plankton
 
 ENTRYPOINT ["/init.sh", "/plankton/simulation.py"]
-
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ To change the rate at which simulation cycles are calculated, increase or decrea
 $ docker run -it dmscid/plankton -d linkam_t95 -p stream -c 0.05
 ```
 
-For long running devices it might be useful to speed up the simulation using the `-e` or `--speed` paramter, which is a factor by which actual time is multiplied to determine the simulated time in each simulation cycle. To run a simulation 10 times faster:
+For long running devices it might be useful to speed up the simulation using the `-e` or `--speed` parameter, which is a factor by which actual time is multiplied to determine the simulated time in each simulation cycle. To run a simulation 10 times faster:
 
 ```
 $ docker run -it dmscid/plankton -d linkam_t95 -p stream -e 10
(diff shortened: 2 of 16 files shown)


MichaelWedel:
Remaining documentation and testing issues have been fixed, ready to be merged from my point of view.

PR #100, opened 2016-09-20
Process Adapter and Device separately

### LINKED ISSUE #97: Simulation should run Adapters and Device separately
At the moment `Simulation` calls the Adapter's process-method once per cycle, which in turn calls the device's process method. Instead, `Simulation` should store device and adapter separately and they should both be processed separately.

This is the first step towards splitting Adapter and Device properly. The interface of Adapter will change, its process-method does not need to know about the simulation delta.

MichaelWedel:
This fixes #97.

Another step towards better separation of Adapter and Device. The `pause` function now only pauses the device, whereas the Adapter should remain responsive (the device just won't do anything as its `process`-method is not called). The new `disconnect_device`-method inhibits calling the `process`-method of the Adapter, effectively making the device appear disconnected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,15 @@ $ ./control.py simulation pause
 $ ./control.py simulation resume
 ```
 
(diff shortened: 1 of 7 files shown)


MikeHart85:
Looks good, as discussed. Merging.

PR #104, opened 2016-09-26
Fix deviation from preliminary EPICS interface

MichaelWedel:
After a request via mail I noticed that the "readback" PVs point to the wrong property as specified by the ICD.


OwenArnold:
@MichaelWedel If the intention is to have the read-back be the target rather than the actuals. This looks good to me.

Issue #95, opened 2016-09-16
Split chopper into Adapter and Device


MichaelWedel:
For the version 1.0 re-design the chopper needs to be split properly into Adapter and Device as described in the design document.

This requires changes to the EPICS adapter and removal of bindings.

PR #106, opened 2016-10-06
Split chopper/restructure package

### LINKED ISSUE #94: Implement new package structure
As part of the re-design for version 1.0, the package structure of plankton needs to change, partly because some concepts will disappear (bindings) and partly because devices will be organised slightly differently.

MichaelWedel:
I'm sorry for the gigantic pull request, but these things turned out to be quite entangled.

This fixes #94 and #95.
# How devices and setups work

Everything. Each device now is a module in `devices`. There are a few rules that devices need to follow so that all the setup and adapter detection works. The following describes different possibilities:

**1. There is only one default setup:**
- Package structure:

```
devices
  device
    __init__.py
```
- The `__init__.py` file must declare one class that inherits from `CanProcess`.
- No setup needs to be passed to `plankton.py`, but it's possible to pass `-s default`.
- This only works for devices that can be default-constructed

**2. There is one device with a few simple setups or more parameters:**
- Package structure:

```
devices
  device
    __init__.py
    device_definition.py # arbitrary name
```
- Content of `device_definition.py`:

```
class SomeDevice(CanProcess):
    # implementation

class SomeAdapter(EpicsAdapter):
    pvs = {
        # blabla
    }
```
- Content of `__init__.py`:

```
from device_definition import SomeDevice
from device_definition import SomeAdapter

setups = dict(
                somesetup=dict(
                    device_type=SomeDevice,
                    parameters={'some_param': 3.0}
                ))
```
- `parameters` is optional if the device is default constructible.
- `planton.py` can now be called with `-s default` and `-s somesetup`

**3. More complex setups**
- Package structure:

```
devices
  device
    setups
      __init__.py
      complex_setup.py
    __init__.py
    device_definition.py
```
- `complex_setup.py` must import at device class as `device_type`. It must declare a dictionary called `parameters` that contains the constructor arguments.

All three possibilities can be used in combinations. The loader first tries the third method, then the second and if all that fails it tries the default (but only if no setup or `default` was specified).
# How adapters work

To find adapters, plankton will first check if `devices/device/adapters/` is a module that can be imported. It imports the adapters it finds there (all adapters need to be imported or defined in the `__init__.py`-file). Then it tries to import adapters from `devices/device`, adds them to the list and checks if one with the proper protocol exists. This happens after the device has been imported using the method described above. If no adapter is specified, any adapter is selected. This depends on the sorting order in a dict, so it's a bit hard to predict.
# Remarks

Later on, a command line option can be added that lets users specify where to look for devices. This is not relevant now as the PyPI-package won't make it into 1.0, so people will need to work with the source anyway.

I have not written any tests for this yet. I've manually tested the different scenarios for the setups and adapters, it seems to work. We need tests for this, but they can be added later.


adapters/epics.py
```diff
@@ -25,75 +25,206 @@
...
+    of the device into SimpleDeviceEpicsAdapter itself, so that it is does not
+    matter whether the specified property in PV exists in the device or the adapter.
+
+    The intention of this design is to keep device classes small and free of
+    protocol specific stuff, such as in the case above where stopping a device
+    via EPICS might involve writing a value to a PV, whereas other protocols may
+    offer an RPC-way of achieving the same thing.
+    """
+    protocol = 'epics'
+    pvs = None
```
OwenArnold, line 140: I assume it was a conscious decision to make this a class attribute? It makes sense given that the PV list is tied to an epics-adapter subclass rather than an epics-adapter sub-class instance.

MichaelWedel: Yes, we discussed this, @MikeHart85 looked at the consequences of this with a minimal example, this is exactly the behavior that we want. Also, we'll never have two devices that use the same `EpicsAdapter`.

OwenArnold:
All tests passing, I'm able to work with both devices via `command.py`. I see no issues that prevent me from merging this.

PR #107, opened 2016-10-07
control.py should not print None return-value

MichaelWedel:
As @OwenArnold pointed out earlier today,`control.py` is slightly confusing in that it also prints `None` when a setter is called that has no return value.

By default the script does not print `None` any more, but for debugging purposes it can be reactivated by supplying the new `-n` or `--print-none` parameter.

It's just a small fix that can be merged without conflicting with the large PR for the refactoring.


OwenArnold:
Tests pass. Works as expected.

OwenArnold:
@MichaelWedel something was stale in my plankton directory. Clearing it got the tests passing again.

PR #110, opened 2016-10-10
Add Device and StateMachineDevice

### LINKED ISSUE #109: Add Device and StateMachineDevice to make implementation easier
As @OwenArnold pointed out, you have to inherit from `CanProcessComposite` or `CanProcess` and do weird stuff to implement a device.

Instead there should be intermediate classes that make it more obvious how to implement a device along the lines of how the adapters work now, with a few member that need to be overridden.

MichaelWedel:
This fixes #109.

### DIFF
--- a/devices/__init__.py
+++ b/devices/__init__.py
@@ -83,9 +83,15 @@ class StateMachineDevice(CanProcessComposite):
         self._initialize_data()
         self._override_data(override_initial_data)
 
+        state_handlers = self._get_final_state_handlers(override_states)
+        initial = self._get_initial_state() if override_initial_state is None else override_initial_state
+
+        if not initial in state_handlers:
+            raise RuntimeError('Initial state \'{}\' is not a valid state.'.format(initial))
+
         self._csm = StateMachine({
-            'initial': self._get_initial_state() if override_initial_state is None else override_initial_state,
-            'states': self._get_final_state_handlers(override_states),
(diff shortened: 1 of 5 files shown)


devices/__init__.py
```diff
@@ -19,7 +19,141 @@
 
 from __future__ import absolute_import
 import importlib
-from core import CanProcess
+from core import CanProcess, CanProcessComposite, StateMachine
+from core.utils import dict_strict_update
+
+
+class Device(CanProcess):
```
OwenArnold, line 26: From the point of view of the device authors, this is going to make far more sense. Even though it's just a name change.

devices/__init__.py
```diff
@@ -19,7 +19,141 @@
...
+    def _get_transition_handlers(self):
+        """
+        Implement this method to return transition handlers for the internal state machine. The keys should be
+        (state, state)-tuples and the values functions that return true if the transition should be triggered.
+
+        :return: A dict-like object containing transition handlers.
+        """
+        raise NotImplementedError('_get_transition_handlers must be implemented in a StateMachineDevice.')
+
+    def _initialize_data(self):
```
OwenArnold, line 127: I was going to say that documentation for this would have been useful. But you beat me to it :)

OwenArnold:
@MichaelWedel. The code changes thus far look good to me.

PR #113, opened 2016-10-10
Trying to fix problems in tests

MichaelWedel:
The tests fail on the build servers for some reason, it looks like part of the problem are the new adapters.


OwenArnold:
@MichaelWedel Thanks for the fix.

PR #114, opened 2016-10-10
Make device argument positional and mandatory

### LINKED ISSUE #111: Make device positional mandatory argument
There should not be a default device, thus the argument should be positional.

MichaelWedel:
This fixes #111.

For a long time we had chopper as the default device, but that seems a bit arbitrary now that the framework is specifically targeting devices in general.

The device argument is now mandatory and thus converted into a positional argument. The `-d`-switch has been removed.

**Note: This is based off branch `fix_tests`, please merge this afterwards.**


OwenArnold:
Works as I'd hoped it would.

PR #115, opened 2016-10-10
Make control.py easier to use

### LINKED ISSUE #112: Make control.py easier to use
The control.py script does not need the -l and -a flags, instead it should just cleverly handle the absence of the parameters and do the right thing.

MichaelWedel:
This fixes #112.

The two options to list all objects and to list an object's API are removed. Instead it is now enough to supply no object argument to get all objects and no member argument to get all members.

**Note: This is branched off `fix_tests`, please merge this branch afterwards.**


OwenArnold:
Works as advertised

Issue #89, opened 2016-09-08
Update documentation once Communication Redesign completed


MikeHart85:
The front-facing `README.md` will need to be updated once design changes [discussed here](https://github.com/DMSC-Instrument-Data/documents/blob/master/design/plankton/CommunicationLayerDesign.md) are applied.

Perhaps this would also be a good time to split the documentation up a bit by topic, put most of the content in `/docs/`, and make the front-facing README leaner with links to the various topics. 

A concise overview with links to more tends to make for a more accessible entry point... Screenshots would also be a great addition, I think.

OwenArnold:
@MikeHart85 / @MichaelWedel I think this is somehow tied to the #108. But we want all documentation to look good for the Release 1.0

MichaelWedel:
Can this be closed now?

OwenArnold:
@MichaelWedel I'm waiting on @MikeHart85 to split the Framework Details off the front README. I'll do it if nothing gets pushed in the next couple of hours.

PR #116, opened 2016-10-12
Features section in README

OwenArnold:
The README is actually very good. But it's lacking what Plankton will do for the reader. I've tried to add a section for that.

Once this is in place @MikeHart85 should be able to take the details out of the front facing README as described here: #89


OwenArnold:
@MikeHart85 and I are going to refine this, but I thought that @MichaelWedel should actually merge the changes.

MichaelWedel:
I looked through the changes so far, it looks very good. This is a really nice section which briefly explains everything Plankton does.

MikeHart85:
Seems like a good place to introduce the "Interface" term that we discussed for the redesign.

An `Adapter` is what Plankton as a framework provides to facilitate communicating over a _transfer_ protocol (EPICS, TCPStream, etc).
An `Interface` is what a device developer implements, using an `Adapter`, to capture the _device_ protocol.

I've only updated the code to reflect this where the user would see it / where it was required to keep things working. So mostly inside the `devices` directory, and anything that had to change as a consequence.

We should still do a complete pass over the code to make sure it is consistent throughout. But that can wait till after `1.0`. I wanted to get this part in before since it will affect the devices / what the users see.

PR #117, opened 2016-10-12
Linear approach function moved to core

### LINKED ISSUE #81: Provide certain behavior as separate classes/functions
Many default states for the devices have very similar code, because essentially they are "approach a target value linearly with a given rate of change". There should be exactly one place in the code where this is implemented.

My suggestion would be to create a new module in `core` for this. Maybe even in a sub-module called `behavior` or something along those lines.

MichaelWedel:
This fixes #81.

There is now an additional `core.approaches` module, which contains a function `linear`. Others could be added easily. I tried implementing this as a class first with rate as a parameter to `__init__`, but the problem is that the rate is not always constant (see Linkam).


OwenArnold:
I'll review this. Probably between meetings today.

core/approaches.py
```diff
@@ -0,0 +1,52 @@
...
+    :return: The new variable value.
+    """
+    sign = (target > current) - (target < current)
+
+    if not sign:
+        return current
+
+    new_value = current + sign * rate * dt
+
+    if sign * new_value > sign * target:
```
OwenArnold, line 49: Ok. So the function will automatically prevent overshooting, if say your `dt` is very coarse

MikeHart85:
I'd already started looking at it yesterday, just wanted to test it with IBEX to make sure the Linkam still behaves as expected.

Something is not quite right with it, I'm looking into it now.

OwenArnold (commented):
I'll leave @MikeHart85  to do the rest of review. As expected code sharing is a lot better. [GPA 3.06](https://codeclimate.com/github/DMSC-Instrument-Data/plankton/compare/approach_function) with the changes

MichaelWedel:
Hm, the unit tests still pass for the linkam and the chopper also seemed to work as expected. Let me know if you find out what it is.

MikeHart85:
The issue wasn't related to this change... it was just a typo in the set_limit command for the Linkam ("L[number]" instead of "L1[number]").

Unit tests wouldn't have seen that since they don't issue the commands but call the interface directly... We'll need to set up system tests to catch things like that.

MikeHart85:
All looks good to me now, should be ready to merge?

OwenArnold:
@MikeHart85. Your changes look fine. I say merge it.

MichaelWedel:
Ah okay, I must have missed that. Good thing you discovered it before 1.0 :)

PR #118, opened 2016-10-13
Issue in literal processing

OwenArnold:
This took me a little while to track down. But if you have a device that has a bool, and you try to set it via `control.py`, the convert_type wrapper does not change the literal from a unicode string to a bool and it is then difficult to find out what the problem is in your `Device`.

### DIFF
--- a/control.py
+++ b/control.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+import ast
 import argparse
 from core.control_client import ControlClient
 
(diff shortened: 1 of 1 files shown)


MichaelWedel:
I've come across this as well, but I think I'd prefer a more restrictive solution, i.e. a function that tries to convert the value to `bool` first (and raises `ValueError` if that does not work so that the current chain with `int` and `float` can be invoked. For the record, if you send `0` (False), or `1` (True) it works as it is.

MichaelWedel:
Okay I just read what this actually does and it seems the entire test chain can be replaced by this.

MichaelWedel:
Which you already did...sorry, doing too many things at once.

PR #119, opened 2016-10-13
Remove regex_flags argument from Stream Cmd

MikeHart85:
Consider this PR a suggestion.

The regex flags can be set directly in the regex using the following syntax:

> (?iLmsux)
>    (One or more letters from the set 'i', 'L', 'm', 's', 'u', 'x'.) The group matches the empty string; the letters set the corresponding flags: re.I (ignore case), re.L (locale dependent), re.M (multi-line), re.S (dot matches all), re.U (Unicode dependent), and re.X (verbose), for the entire regular expression. (The flags are described in Module Contents.) This is useful if you wish to include the flags as part of the regular expression, instead of passing a flag argument to the re.compile() function.
>    Note that the (?x) flag changes how the expression is parsed. It should be used first in the expression string, or after one or more whitespace characters. If there are non-whitespace characters before the flag, the results are undefined.

This makes exposing regex_flags as an argument a bit redundant. Removing it makes the semantics of creating a Cmd a bit cleaner. You can just do, for example:

``` python
Cmd('set_rate', '^R1([0-9]+)$', (int,))
```

Without having to name the argument_mappings parameter to get around the regex_flags parameter.

### DIFF
--- a/adapters/stream.py
+++ b/adapters/stream.py
@@ -87,7 +87,7 @@ class StreamServer(asyncore.dispatcher):
 
 
 class Cmd(object):
-    def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
+    def __init__(self, target_method, regex, argument_mappings=None,
                  return_mapping=lambda x: None if x is None else str(x)):
         """
         This is a small helper class that makes it easy to define commands that are parsed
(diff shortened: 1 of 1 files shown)


MichaelWedel:
Interestingly and unexpectedly (at least for me), this suggestion turns out to make adding `Var` semantics as mentioned in #165 and #166 much easier and more consistent. Apologies for taking so long with this, but I did not see enough "pressure" to remove the argument - now I do. Due to the rename and so on, merge conflicts have accumulated, so my suggestion would be to do this in the context of #166.

MikeHart85:
Sounds good! :+1:

PR #120, opened 2016-10-13
Add contribution guide line

### LINKED ISSUE #108: Add contribution guideline document
As described in the design document, a contribution guidelines document is required. It should describe the principles of `Adapter` and `Device` and how to add a new device.

MichaelWedel:
This fixes #108.

I hope the text describes somewhat how to add a device. There is also an `examples` directory, which could be moved to `docs` as well, but I have no preference for either.

To make the example runnable and keep them out of the `devices`-directory, I decided to modify the startup script so that it's possible to start devices from other places.


OwenArnold:
Tested out all examples. Ran through instructions. Read documentation. All looks great.

PR #127, opened 2016-10-19
Added basic flake8 configuration

### LINKED ISSUE #124: Add flake8 job to travis
The package [flake8](https://pypi.python.org/pypi/flake8) helps enforce many of the style guide options (line length, unused variables, etc.).

This should be a part of the travis configuration so that we don't add non-confirming stuff.

MichaelWedel:
This fixes #124.

All PEP8 issues are resolved now, from now on new code is not allowed to introduce new PEP8 problems. I've arbitrarily set the maximum complexity for functions to 12. Only `import_device` has this value, it should be refactored and the limit set to a lower value (maybe 10?), but that felt a bit outside of the scope for this issue (that function has no test yet). The only code I've refactored because there are tests and it was easy was in `StateMachine.__init__`.

We have to take a look at `core.__init__.py` and so on at some point to make sure `__all__` contains everything we intend to make available.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+max-complexity = 12
+exclude = .git, docs
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,14 @@ services:
 python:
   - "2.7"
   - "3.5"
+
 install:
   - pip install -r requirements-dev.txt
   - pip install coveralls
+  - pip install flake8
+
 script:
+  - flake8 .
   - nosetests -w test --with-coverage --cover-package=core,devices
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ python:
 install:
   - pip install -r requirements-dev.txt
   - pip install coveralls
-  - pip install flake8
 
 script:
   - flake8 .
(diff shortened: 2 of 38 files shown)


MichaelWedel:
450 issues...but 350 or so are long lines, all in all this shouldn't be too hard to fix and maybe it's a good time to do it now before we begin "serious" work for the next release.

MichaelWedel:
I tried using `autopep8` earlier today, but it only fixed something like 100 issues or so, even in very aggressive mode and some in weird ways (many line breaks etc.). Maybe the best way is to do it manually.

MichaelWedel:
Interesting according to codacy, I overlooked and misunderstood `__all__` completely.

MikeHart85:
I don't think I like the 79 char line limit. 

It forces even simple things to be split across multiple lines. This harms readability rather than improving it in my opinion.

Even PyCharm (which is very picky) defaults to 120, which I think would be much more reasonable.

MichaelWedel:
We've had discussions around this in Mantid as well, clang-format also enforces an 80 character line length limit. In the beginning it looked a little odd, but that feeling went away quickly (at least for me).

Some cases of long lines are statements like `a = something if something else something`, maybe it's worthwhile in some of these cases to refactor into a smaller function.

Other Python projects seem to adhere to this standard and it is specifically mentioned in the official style guide: https://www.python.org/dev/peps/pep-0008/#maximum-line-length
It mentions allowing code to exceed that limit to maximum 100, but still applying the stricter limit to comments.

My vote on this is to stick to what the standard suggests.

PR #137, opened 2016-10-26
Turn plankton into a package

### LINKED ISSUE #125: Transform plankton into an installable package
As we have discussed a few times already, we should re-arrange the code in a way that makes Plankton distributable as a package installable via pip (and produce such a package).

Like other packages such as `nose` we could ship `plankton` as an executable so that it can be run directly from the command line via `plankton chopper`. On the other hand, it should be possible to use plankton in python scripts:

``` python
from plankton.core import StateMachine
from plankton.core import ControlClient
```

Related to this I suggest we keep the devices in that distribution. They might not all be useful to everybody, but it would be easier to write scripts that involve these devices but not other parts of Plankton, for example like this:

``` python
from plankton.devices.chopper import SimulatedChopper

# Do something interesting with the chopper
```

This would also simplify creation of the docker images even further.

In my opionien #10 is a prerequisite for this, so that users can find documentation online easily.

MichaelWedel:
This fixes #125.

In this form the library would be ready for packaging with a `setup.py` file. Everything in the `plankton` (to be renamed) directory would be shipped. I've moved `plankton.py` into `plankton/scripts` and renamed them. Instead of scripts they are now functions `run_simulation` and `control_simulation`. Both have an optional parameter to pass an argument list instead of `sys.argv[1:]`, this is useful for putting that code under test as well later on.

In addition to `plankton.py` and `plankton-control.py` in the top level directory, which are just for development purposes, the upcoming `setup.py` file will use the existing `setuptools`-logic to generate two "executables", one `plankton` (to be renamed) and one `plankton-control` (also to be renamed) that contain generated code to run the appropriate functions from the appropriate modules. The only thing I'm not happy with is that `plankton.py` and `plankton/__main__.py` are exact copies. Any ideas?

Like for other modules that contain a `__main__.py`, it is now also possible to run the module directly like this:

```
python -m plankton chopper
```

On of the things the `-m`-flag does is that it modifies the PYTHONPATH, it adds `.`. This is really useful in conjunction with `-k`, because you don't need to explicitly add `-a` to add directories to the path.


MichaelWedel:
Documentation has not been touched so far, I'll do that once the structure is acceptable.

MikeHart85:
`setup.py` will not be added as part of this pull request?

What is the purpose of the `plankton/plankton` directory? Is it just for making it easier to package? Could it be done without it? For that matter, I'm surprised things like `doc`, `tests` and `LICENSE` would not be included. 

Wouldn't it be easier, and more consistent, to just deliver everything (except for obvious exclusions like `.git` and the like)?

MichaelWedel:
LICENSE and README probably have to move, yes.

The packages I looked at (nose2, sphinx, a few others) all have this structure. `setup.py` at top level and then the actual package in a sub-directory. `tests` seem to be excluded in many cases, probably because it's meant as a distribution for end-users?

MichaelWedel:
And to answer the first question, no, I've opened a new issue #138.

PR #139, opened 2016-10-27
Remove additional space in coding comment

### LINKED ISSUE #136: Magic coding comment has an errant space
The magic encoding tag at the top of every file:

```
#  -*- coding: utf-8 -*-
```

Has two spaces between `#` and `-*-`. It should be just one space.

This is an outrage and an abomination, and must be addressed before it leads to some sort of calamity.

MichaelWedel:
This fixes #136.

I don't think this requires any further description.


MikeHart85:
I'm glad we can close this shameful chapter in our development history. :+1:

Issue #135, opened 2016-10-26
Standardize Exception Handling


MikeHart85:
Currently, exception handling is very ad hoc throughout the system. 

The following exception classes exist:
- `StateMachineException` (statemachine.py)
- `RemoteException` (control_client.py)
- `ProtocolException` (control_client.py)

The latter two are probably fine as-is, since the control client is arguably its own application/library and those two exception types work well to capture and differentiate what can go wrong during its usage.

The simulation itself, however, needs better exception types and usage practices. Currently, only the StateMachineException exists as a distinct exception type. Any time an exception needs to be raised outside of that, standard classes like RuntimeError are used. _Even inside the StateMachine, sometimes a RuntimeError is raised instead of StateMachineException._

This makes catching and handling exceptions difficult and, consequently, we don't do that very well either.

Ideas to consider/discuss:
- We should keep in mind that there are two distinct usage scenarios for the simulation: Command-line and as an imported library.
- Should we have a `core/exceptions.py` or similar to define all exceptions used throughout the system?
- Should we avoid allowing anything apart from exceptions we define to escape our code as a matter of principle? Raising standard exceptions within the system would be fine as long as they are handled before reaching the user under this model.
- One could argue that exceptions like AttributeError, ValueError, NotImplementedError, etc are descriptive enough and don't need to be wrapped in our own type / would be better as they are.
- Any of the above exceptions are meaningless if they come from too deep in our code. Getting an `IndexError` makes sense if you as the user provided a bad index. It doesn't make sense if you called a function that doesn't even take parameters, but something went wrong deep inside the core code (like, say, `pcaspy` couldn't be imported).
- RuntimeError, however, is so generic that I don't think it should ever escape our code (or perhaps even ever be raised?).
- On the other hand, we could consider not using our own exception types at all (including getting rid of StateMachineException) and just using standard exceptions for everything (in which case RuntimeError might be fine if there isn't anything more appropriate). This might make handling exceptions awkward though.
- When run from the command-line, we should have some high-level exception handling to make sure no exceptions escape to become traces. Errors should be displayed in a human readable way for the user, and ideally be specific enough to enable finding a solution easily.
- Interrupts like CTRL-C should also be handled gracefully.

This issue can be used to discuss how to best solve this.

This issue should probably be addressed after dealing with #125, #132 and #133.

MichaelWedel:
[...]

Issue #134, opened 2016-10-26
Fail more gracefully when optional dependencies are missing


MikeHart85:
PCASPy and EPICS are optional dependencies. When they are missing, but the user attempts to use something that depends on them, the system should fail gracefully and inform the user what went wrong and why.

Right now, the exceptions thrown when attempting to run an EPICS device without proper dependencies are not clear as to what failed or what can be done about it, and can even be misleading.

Related to #135.

MichaelWedel:
As per your suggestion I will make a start by raising more meaningful `PlanktonException`s with better text.

PR #142, opened 2016-11-02
Simulation can construct its own ControlServer

### LINKED ISSUE #96: SimulationEnvironment should construct its own ControlServer
At the moment this is accessible from the outside, but ultimately the environment should own the server and construct it.

MichaelWedel:
Simulation now accepts different input, most importantly a `host:port` string from which a ControlServer is constructed. For consistency this is also available via the property setter.

For better control over when the server starts listening, it has to be started explicitly now, either via calling `start_server` or via a new parameter of the constructor called start.

A few more tests have been added around this. Some of the exceptions may be seen by users, those have been converted to `PlanktonException`s.

Fixes #96.

### DIFF
--- a/plankton/core/control_server.py
+++ b/plankton/core/control_server.py
@@ -164,11 +164,13 @@ class ExposedObjectCollection(ExposedObject):
 
 
 class ControlServer(object):
-    def __init__(self, object_map=None, host='127.0.0.1', port='10000'):
+    def __init__(self, object_map=None, host='127.0.0.1', port='10000', start=False):
         """
-        This server opens a ZMQ REP-socket at the given host and port. It constructs an
-        ExposedObjectCollection from the supplied name: object-dictionary and uses that
-        as a handler for JSON-RPC requests. If it is an instance of ExposedObject,
+        This server opens a ZMQ REP-socket at the given host and port when start_server
+        is called. If the start-parameter is set to True, this happens directly on construction.
+
+        The server constructs an ExposedObjectCollection from the supplied name: object-dictionary
+        and uses that as a handler for JSON-RPC requests. If it is an instance of ExposedObject,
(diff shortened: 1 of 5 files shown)


plankton/core/simulation.py
```diff
@@ -83,6 +86,28 @@ def __init__(self, device, adapter, control_server=None):
...
+        self.control_server = control_server
+
+    def _create_control_server(self, control_server):
+        if control_server is None or isinstance(control_server, ControlServer):
+            return control_server
+        else:
+            arguments = control_server.split(':')
+
+            if len(arguments) not in (1, 2):
+                raise RuntimeError(
```
MichaelWedel: @MikeHart85 this is one of the errors I meant in my comment in #135. The message is probably okay already, but it needs to be a distinct type so that the message can be displayed in a good way. Tomorrow I'll try to make some progress on #134, where I've already worked on the display part.

plankton/core/control_server.py
```diff
@@ -193,16 +196,37 @@ def __init__(self, object_map=None, host='127.0.0.1', port='10000'):
         self.port = port
```
MichaelWedel, line 205: The `RuntimeError` above this (you have to expand the diff view a bit to see it) is probably also of the category of potentially user facing errors.

MikeHart85: Yes, that message is meant for the user (to let them know they've set the `-r` option incorrectly, or that something went wrong with it). Should probably become a PlanktonException?

MichaelWedel:
The problem in the unit tests was that `ControlServer` was mocked, but the return-value was unspecified. So the call was on a different instance of whatever the control server mock returned.

This is fixed now.

MikeHart85:
Looks good, ready to merge.

PR #147, opened 2016-11-04
Better error messages on command line

### LINKED ISSUE #134: Fail more gracefully when optional dependencies are missing

MichaelWedel:
This fixes #134.

A new exception module has been added to `core`, which contains one type at the moment `PlanktonException`. It is raised in a few common places where we expect things to go wrong if `pcaspy` is not installed or the devices are somehow not in the path.

For better display `run_simulation` is now a wrapper that watches for these exceptions, catches them and displays them in a nice way. A wrapper function is used to still be able to test the actual exception throwing once we have a test for this. All other exceptions are not handled, as I explained in another issue my opinion is that these unexpected exceptions are errors on our side and should be reported (and then fixed).


plankton/adapters/epics.py
```diff
@@ -23,10 +23,19 @@
...
+from plankton.core.exceptions import PlanktonException
+
+# pcaspy might not be available. To make EPICS-based adapters show up anyway,
+# dummy types are created in this case and the failure is postponed to
+# runtime, where a more appropriate PlanktonException can be raised.
+try:
+    from pcaspy import Driver, SimpleServer
+except ImportError:
+    Driver = type('Driver', (object,), {})
+    SimpleServer = type('SimpleServer', (object,), {})
```
MikeHart85: Can we avoid this? Doesn't feel right to just patch it to work-until-it-doesn't like this.

What is meant by "make EPICS-based adapters show up"? When using `--list-protocols`? Should protocols that aren't available due to missing dependencies really show up  as available?

Ideally, I think the ImportError should be allowed to propagate rather than being handled directly here. Since we import devices/interfaces dynamically, the part of the code that triggered this module (epics.py) being imported should watch for the ImportError and react as appropriate. This would be a more general solution, as we will likely be adding more adapters, which may well have their own non-standard module requirements.

plankton.py
```diff
@@ -21,4 +21,4 @@
 from plankton.scripts.run import run_simulation
 
 if __name__ == '__main__':
-    run_simulation()
+    print(run_simulation() or '')
```
MikeHart85: Shouldn't `__main__.py` change the same way (if this stays like this; see comments below first)?

plankton/scripts/run.py
```diff
@@ -61,23 +62,23 @@
...
 
 
-def run_simulation(argument_list=None):
+def do_run_simulation(argument_list=None):
     arguments = parser.parse_args(argument_list or sys.argv[1:])
 
     if arguments.version:
-        print(__version__)
-        return
+        return __version__
```
MikeHart85: Why the shift to returning rather than printing? 

I don't really see anything wrong with printing here since the scripts directory is for command line interfaces. `control.py` still prints as well.

plankton/scripts/run.py
```diff
@@ -61,23 +62,23 @@
...
 
     if arguments.add_path is not None:
         sys.path.append(os.path.abspath(arguments.add_path))
 
     if not arguments.device:
-        print('Please specify a device to simulate.')
-        print('The following devices are available:')
+        ret = ['Please specify a device to simulate.'
+               'The following devices are available:']
```
MikeHart85: I think this is missing a comma?

MikeHart85 (commented):
Trying out review system

plankton/scripts/run.py
MichaelWedel: Yes, I think it is.

plankton/scripts/run.py
MichaelWedel: The advantage of returning is that we can have tests for the function and can check what it will actually print. Catching stdout just complicates the tests.

plankton.py
MichaelWedel: Yes that's absolutely right.

plankton/adapters/epics.py
MichaelWedel: The downside of handling the import error in the client code is that every single client has to do it. If there are 10 devices with an EPICS interface, all 10 will have to `try`-`import`-`except`, this way it's only there once.

I think when using `--list-protocols` all adapters that are in the selected device module should be displayed, because otherwise users might not know that they exist. When they see one they want to use and try it, they will see the error describing what they have to do.

If the error is handled inside `import_adapter` the user will never be able to try and load the adapter and again never know it exists.

MichaelWedel:
I'm trying to make the "replace with dummy types if import fails"-part a bit less ugly and a bit more general and re-useable. I still think it's the right thing do to, maybe it looks less scary if done in a slightly better way.

plankton/adapters/epics.py
```diff
@@ -23,10 +23,16 @@
...
 
-from plankton.core.utils import seconds_since
+from plankton.core.utils import seconds_since, From
+from plankton.core.exceptions import PlanktonException, StubAccessException
+
+# pcaspy might not be available. To make EPICS-based adapters show up
+# in the listed adapters anyway dummy types are created in this case
+# and the failure is postponed to runtime, where a more appropriate
+# PlanktonException can be raised.
+Driver, SimpleServer = From('pcaspy').import_or_stub('Driver', 'SimpleServer')
```
MichaelWedel: What do you think about this approach? It's still delaying the ImportError, but it's completely general now and can be used wherever we have this kind of optional dependency. First I had a function `def import_or_stub(module, *names)`, but it looked weird. This is not overly pretty either, but it reads in a similar way as normal import statements.

plankton/core/utils.py
```diff
@@ -115,3 +119,66 @@ def seconds_since(start):
...
+        except ImportError:
+            def create_getattr_function(name):
+                def failing_init(obj, *args, **kwargs):
+                    raise StubAccessException(
+                        'You are trying to instantiate a stub-type that '
+                        'replaces \'{}\' from module \'{}\''.format(name, self._module))
+
+                return failing_init
+
+            return tuple(type(name, (object,), {'__init__': create_getattr_function(name)})
```
MichaelWedel: Here the stub-types are generated with an `__init__`-method that always raises a `StubAccessException` which is easy to catch and distinguish from other exceptions - it's really nice to have different exceptions for these things.

If the module exists and you try to import something from it that does not exist, you get an `AttributeError`, so it's "safe" in that way.

plankton/scripts/run.py
MichaelWedel: I've thought further about this. It looks like we're on a path towards separating the construction of the simulation from running it as well. This can be useful for running simulations in a different context (you mentioned a service in one of the issues). The context might not support printing, but it could still start from using this function and evaluating its return value.

plankton/adapters/epics.py
```diff
@@ -23,10 +23,23 @@
...
+# and the failure is postponed to runtime, where a more appropriate
+# PlanktonException can be raised.
+missing_pcaspy_exception = PlanktonException(
+    'In order to use EPICS-interfaces, pcaspy must be installed:\n'
+    '\tpip install pcaspy\n'
+    'A fully working installation of EPICS-base is required for this package. '
+    'Please refer to the documentation for advice.')
+
+Driver, SimpleServer = FromOptionalDependency(
+    'pcaspy', missing_pcaspy_exception).do_import('Driver', 'SimpleServer')
```
MichaelWedel, line 42: I'm happier with the way this looks now. The `FromOptionalDependency` can be used for any optional module now and the exception that is raised when any of the replacement types are instantiated is fully controllable, in this case a `PlanktonException` is explicitly created, but it would be enough to pass the text - the class would create a `PlanktonException` from it by default.

What I do not like is the `do_import`, but you can't use `import`, so that's somehow the closest thing I was able to come up with.

MichaelWedel:
I've tried to address the problem pointed out the in the review. Besides fixing the obvious bugs, the "optional import" functionality is now better defined and can be used in a general way. What do you think about this?

This way `ImportErrors` of optional dependencies are handled in one place and whenever we have a new optional dependency there's no need to worry about this.

plankton/core/utils.py
```diff
@@ -115,3 +120,85 @@ def seconds_since(start):
...
+            return tuple(getattr(module_object, name) for name in names)
+        except ImportError:
+            def create_getattr_function(name):
+                def failing_init(obj, *args, **kwargs):
+                    raise self._exception
+
+                return failing_init
+
+            return tuple(type(name, (object,), {'__init__': create_getattr_function(name)})
+                         for name in names)
```
MikeHart85: Consider simplifying to:

``` python
def failing_init(*args, **kwargs):
    raise self._exception

return tuple(type(name, (object,), {'__init__': failing_init}) for name in names)
```

Would that work the same way?

plankton/core/utils.py
```diff
@@ -115,3 +120,85 @@ def seconds_since(start):
...
+            exception = 'The optional dependency \'{}\' is required for the ' \
+                        'functionality you tried to use.'.format(self._module)
+
+        if isinstance(exception, string_types):
+            exception = PlanktonException(exception)
+
+        if exception is not None and not isinstance(exception, BaseException):
+            raise RuntimeError(
+                'The exception parameter has to be either a string or a an instance of an '
+                'exception type (derived from BaseException).')
```
MikeHart85, line 179: Should this really be a `RuntimeError`? Or a `PlanktonException`? Or perhaps an `ImportError`?

Nitpick: `exception` cannot be None here since we checked and did something about it before.

test/test_utils.py
```diff
@@ -179,3 +180,38 @@ def test_seconds_since_none(self, datetime_mock):
         datetime_mock.now.return_value = datetime(2016, 9, 1, 2, 0)
 
         self.assertRaises(TypeError, seconds_since, None)
+
+
+class TestFromOptionalDependency(unittest.TestCase):
+    def test_existing_module_works(self):
+        a, = FromOptionalDependency('time').do_import('sleep')
```
MikeHart85: Would it be better to just return `sleep` rather than `(sleep,)` in this case?

plankton/adapters/epics.py
MikeHart85: It looks great, I like it much better than before! :+1: 

I still wish it was even more centralized / transparent, so that any and all exceptions when importing an adapter are automatically caught, without having to do anything special in the adapter code. But I can't really think of a complete solution that would be better than this one. 

Perhaps the adapter modules themselves could be wrapped? So that a fake `EpicsAdapter` can be provided, if importing the real one fails, which throws the previously caught exception when you try to use it? On the other hand, that would require code which imports or creates a fake `EpicsAdapter` outside of `adapter/epics.py`, which would make adapters less drag-and-droppable.

Anyway, definitely not something I would worry about in this PR... maybe someday when we have nothing better to do.

plankton/scripts/run.py
MikeHart85: I'm really not convinced on this.

`plankton.py`, `__main__.py` and `plankton-control.py` are really just stubs that we use as cross-platform symlinks. They shouldn't contain any real functionality.

Both `scripts/run.py` and `scripts/control.py`, on the other hand, are really console applications rather than modules. They define an `argparser` and use it to parse command line arguments, initializing and running the simulation based on that. It makes sense for them to print to the command line.

It doesn't make sense to force them to only return strings, because this limits how and when they can output to the command line: at the end or when something goes wrong. How would we handle a `--verbose` option? Outputting log messages to the command line?

I also don't think writing unit tests for them is necessary / wouldn't really make sense. They're applications, not units.

plankton/core/utils.py
MichaelWedel: True, exception could be `None` in an earlier version of this, but now it can't. Fixing it. But I think it should be a `RuntimeError`, because giving anything else than an exception type or a string is a programming error and it is only something to notify developers, not users.

plankton/core/utils.py
MichaelWedel: Yes, again a leftover from before, where the exception was only raised when an attribute was accessed, but even for my taste that was too late :)

test/test_utils.py
MichaelWedel: I've thought about it, but I opted for consistently returning a `tuple` in order to not create special cases. Maybe it's better to allow the special case of 1.

MichaelWedel:
For some reason I can't reply to the script thing, so I'm replying here.

> They define an argparser and use it to parse command line arguments, initializing and running the simulation based on that. It makes sense for them to print to the command line.

This totally convinces me. Also, it's not in the scope of this issue, so probably better not to touch it anyway.

MikeHart85:
All looks good to me, ready to merge.

PR #148, opened 2016-11-08
Make devices self-documenting

### LINKED ISSUE #141: Make device Interfaces self-documenting
Following some user feedback acquired by @OwenArnold that pointed out device `Interfaces` could be self-documenting with regard to the PVs/commands they expose.

This could be implemented in the respective Adapter, potentially using the docstrings of the exposed properties/methods. These can be extracted using the `inspect`-module.

MichaelWedel:
This fixes #141.

This is my suggestion for "self-documenting" devices. `Adapter` now has an additional property `documentation` which can be overridden in sub-classes. For now I've decided against adding stuff to the class doc-string for several reasons, the most important being that the type of the interface is modified on construction and so it depends on the device and/or configuration (PV-prefix, host/port, ...).

Both `EpicsAdapter` and `StreamAdapter` auto-generate command lists from the `PV`/`Cmd`-objects. These now have `doc`-parameter that can be used to document a command explicitly. If `doc` is not supplied, the docstring of the wrapped/exposed property or method is used, no matter if it's on the device or the interface.

For both adapters, the interface class doc-string is included in the documentation, in the `StreamAdapter`-case also information about host, port, terminators.

To query a device before running it there is a new script parameter:

```
$ ./plankton.py -i chopper -- -p Chopper1:
```

It's also exposed via the `ControlServer`, as the `device_documentation`-property of the `simulation`-object. It made more sense to me to put it there that in the device, because it's related to the interface, not the device.

@dmichel76, is this along the lines you had in mind?

I'll leave the `In progress`-label on for a while so we can discuss.

### DIFF
--- a/plankton/adapters/__init__.py
+++ b/plankton/adapters/__init__.py
@@ -175,5 +175,10 @@ class ForwardMethod(object):
         self._target = target
         self._method = method
 
+        try:
+            self.__doc__ = getattr(self._target, self._method).__doc__
+        except AttributeError:
+            pass
+
     def __call__(self, *args, **kwargs):
         return getattr(self._target, self._method)(*args, **kwargs)
--- a/plankton/adapters/__init__.py
+++ b/plankton/adapters/__init__.py
@@ -18,6 +18,7 @@
 # *********************************************************************
 
 import importlib
+import inspect
 from ..core.exceptions import PlanktonException
 
 
@@ -27,6 +28,10 @@ class Adapter(object):
     def __init__(self, device, arguments=None):
         super(Adapter, self).__init__()
         self._device = device
+
+    @property
+    def documentation(self):
+        return inspect.getdoc(self) or ''
 
     def start_server(self):
         pass
--- a/plankton/adapters/__init__.py
+++ b/plankton/adapters/__init__.py
@@ -113,7 +113,7 @@ def import_adapter(device_name, protocol_name, device_package='devices'):
 
 
 class ForwardProperty(object):
-    def __init__(self, target_member, property_name):
+    def __init__(self, target_member, property_name, instance=None):
         """
         This is a small helper class that can be used to act as
         a forwarding property to relay property setting/getting
@@ -128,7 +128,7 @@ class ForwardProperty(object):
 
             a.forward = 10 # equivalent to a._b.baz = 10
 
-        Note that this modifies the type Baz. Usage must thus be
+        Note that this modifies the type Foo. Usage must thus be
         limited to cases where this type modification is
         acceptable.
 
@@ -137,6 +137,12 @@ class ForwardProperty(object):
         """
         self._target_member = target_member
         self._prop = property_name
+
+        try:
+            self.__doc__ = getattr(type(getattr(instance, self._target_member)),
+                                   self._prop).__doc__
+        except AttributeError:
+            pass
 
     def __get__(self, instance, type=None):
         """
@@ -147,7 +153,10 @@ class ForwardProperty(object):
         :param type: Type.
         :return: Attribute value of member property.
         """
-        return getattr(getattr(instance, self._target_member), self._prop)
+        if instance is not None:
+            return getattr(getattr(instance, self._target_member), self._prop)
+
+        return self
 
     def __set__(self, instance, value):
         """
@@ -157,7 +166,8 @@ class ForwardProperty(object):
         :param instance: Instance of type.
         :param value: Type.
         """
-        setattr(getattr(instance, self._target_member), self._prop, value)
+        if instance is not None:
+            setattr(getattr(instance, self._target_member), self._prop, value)
 
 
 class ForwardMethod(object):
--- a/plankton/adapters/__init__.py
+++ b/plankton/adapters/__init__.py
@@ -133,16 +133,20 @@ class ForwardProperty(object):
         acceptable.
 
         :param target_member: Target member to forward to.
-        :param prop: Property of target to access.
+        :param property_name: Property of target to access.
+        :param instance: Object from which to obtain target_member for the purpose of extracting
+        the docstring of the property identified by property_name. If it doesn't exist on the type,
+        of target_member, the docstring is not copied.
         """
         self._target_member = target_member
         self._prop = property_name
 
-        try:
-            self.__doc__ = getattr(type(getattr(instance, self._target_member)),
-                                   self._prop).__doc__
-        except AttributeError:
-            pass
+        # Extract docstring from the property that's being forwarded.
+        # The property exists in the type of the specified target_member of instance,
+        # so getattr must be called on the type, not object, otherwise the
+        # docstring of the returned value would be stored.
+        self.__doc__ = getattr(type(getattr(instance, self._target_member)),
+                               self._prop, None).__doc__
 
     def __get__(self, instance, type=None):
         """
@@ -166,8 +170,8 @@ class ForwardProperty(object):
         :param instance: Instance of type.
         :param value: Type.
         """
-        if instance is not None:
-            setattr(getattr(instance, self._target_member), self._prop, value)
+
+        setattr(getattr(instance, self._target_member), self._prop, value)
 
 
 class ForwardMethod(object):
@@ -175,10 +179,7 @@ class ForwardMethod(object):
         self._target = target
         self._method = method
 
-        try:
-            self.__doc__ = getattr(self._target, self._method).__doc__
-        except AttributeError:
-            pass
+        self.__doc__ = getattr(self._target, self._method).__doc__
 
     def __call__(self, *args, **kwargs):
         return getattr(self._target, self._method)(*args, **kwargs)
--- a/plankton/adapters/epics.py
+++ b/plankton/adapters/epics.py
@@ -25,9 +25,8 @@ import inspect
 
 from . import Adapter, ForwardProperty
 from six import iteritems
-import textwrap
 
-from plankton.core.utils import seconds_since, FromOptionalDependency
+from plankton.core.utils import seconds_since, FromOptionalDependency, format_doc_text
 from plankton.core.exceptions import PlanktonException
 
 # pcaspy might not be available. To make EPICS-based adapters show up
@@ -171,13 +170,10 @@ class EpicsAdapter(Adapter):
     def documentation(self):
         pv_template = '{} ({}{}):\n{}'
 
-        def wrapped(text):
-            return textwrap.fill(text, width=80, initial_indent='    ', subsequent_indent='    ')
-
         pvs = [pv_template.format(
             self._options.prefix + name,
             pv.config.get('type', 'float'), ', read only' if pv.read_only else '',
-            wrapped(pv.doc or inspect.getdoc(getattr(type(self), pv.property)) or ''))
+            format_doc_text(pv.doc or inspect.getdoc(getattr(type(self), pv.property)) or ''))
                for name, pv in self.pvs.items()]
 
         return '\n\n'.join([inspect.getdoc(self), 'PVs\n==='] + pvs)
--- a/plankton/adapters/epics.py
+++ b/plankton/adapters/epics.py
@@ -21,9 +21,11 @@ from __future__ import print_function
 
 from argparse import ArgumentParser
 from datetime import datetime
+import inspect
 
 from . import Adapter, ForwardProperty
 from six import iteritems
+import textwrap
 
 from plankton.core.utils import seconds_since, FromOptionalDependency
 from plankton.core.exceptions import PlanktonException
@@ -43,7 +45,7 @@ Driver, SimpleServer = FromOptionalDependency(
 
 
 class PV(object):
-    def __init__(self, target_property, poll_interval=1.0, read_only=False, **kwargs):
+    def __init__(self, target_property, poll_interval=1.0, read_only=False, doc=None, **kwargs):
         """
         The PV-class is used to declare the EPICS-interface exposed by a sub-class of
         EpicsAdapter. The target_property argument specifies which property of the adapter
@@ -56,11 +58,13 @@ class PV(object):
         :param target_property: Property of the adapter to expose.
         :param poll_interval: Update interval of the PV.
         :param read_only: Should be True if the PV is read only.
+        :param doc: Description of the PV. If not supplied, docstring of mapped property is used.
         :param kwargs: Arguments forwarded into pcaspy pvdb-dict.
         """
         self.property = target_property
         self.read_only = read_only
         self.poll_interval = poll_interval
+        self.doc = doc
         self.config = kwargs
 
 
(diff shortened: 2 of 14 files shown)


MichaelWedel:
Somehow the travis build was not triggered, I'll try again by closing and re-opening.

plankton/adapters/__init__.py
```diff
@@ -133,6 +138,12 @@ def __init__(self, target_member, property_name):
         self._target_member = target_member
         self._prop = property_name
 
+        try:
+            self.__doc__ = getattr(type(getattr(instance, self._target_member)),
+                                   self._prop).__doc__
+        except AttributeError:
+            pass
```
MikeHart85: Would it be better to check `instance` for None and still treat `AttributeError`s as errors?

plankton/adapters/__init__.py
```diff
@@ -152,13 +166,19 @@ def __set__(self, instance, value):
         :param instance: Instance of type.
         :param value: Type.
         """
-        setattr(getattr(instance, self._target_member), self._prop, value)
+        if instance is not None:
+            setattr(getattr(instance, self._target_member), self._prop, value)
```
MikeHart85: I don't think `instance` can be `None` in `__set__`... That would imply a scenario like this:

``` python
class Foo(object):
    bar = ForwardProperty(...)

Foo.bar = 'hello world'
```

But, in this case, `bar` itself (the `ForwardProperty` instance) is replaced by 'hello world', rather than the `ForwardProperty.__set__` being called.

plankton/adapters/__init__.py
```diff
@@ -152,13 +166,19 @@ def __set__(self, instance, value):
...
 
 class ForwardMethod(object):
     def __init__(self, target, method):
         self._target = target
         self._method = method
 
+        try:
+            self.__doc__ = getattr(self._target, self._method).__doc__
+        except AttributeError:
+            pass
```
MikeHart85: Is this try catch necessary?

Assuming this is for the case that `_method` has no docstring... doesn't `__doc__` default to `None`, making the assignment succeed?

And if `_method` doesn't actually exist, shouldn't this be treated as an error rather than silently ignored?

plankton/adapters/epics.py
```diff
@@ -163,6 +166,19 @@ def __init__(self, device, arguments=None):
...
 
+    @property
+    def documentation(self):
+        pv_template = '{} ({}{}):\n{}'
+
+        pvs = [pv_template.format(
+            self._options.prefix + name,
+            pv.config.get('type', 'float'), ', read only' if pv.read_only else '',
+            format_doc_text(pv.doc or inspect.getdoc(getattr(type(self), pv.property)) or ''))
+               for name, pv in self.pvs.items()]
```
MikeHart85: That's a really long and convoluted list comprehension. Consider unwrapping into for-loop for better readability / maintainability.

plankton/adapters/__init__.py
```diff
@@ -108,7 +113,7 @@ def import_adapter(device_name, protocol_name, device_package='devices'):
 
 
 class ForwardProperty(object):
-    def __init__(self, target_member, property_name):
+    def __init__(self, target_member, property_name, instance=None):
```
MikeHart85, line 116: Missing `:param:` description for `instance`.

plankton/adapters/__init__.py
```diff
@@ -133,6 +138,12 @@ def __init__(self, target_member, property_name):
         self._target_member = target_member
         self._prop = property_name
 
+        try:
+            self.__doc__ = getattr(type(getattr(instance, self._target_member)),
+                                   self._prop).__doc__
```
MikeHart85: I'm having a hard time understanding what this is actually doing, might benefit from comments and/or breaking down into smaller steps, in addition to description of instance parameter above.

plankton/adapters/epics.py
```diff
@@ -179,7 +195,7 @@ def _create_properties(self, pvs):
                 if prop not in dir(self._device):
                     raise AttributeError('Can not find property \''
                                          + prop + '\' in device or interface.')
-                setattr(type(self), prop, ForwardProperty('_device', prop))
+                setattr(type(self), prop, ForwardProperty('_device', prop, instance=self))
```
MikeHart85, line 203: Already mentioned in other comments, but here again, as a user of `ForwardProperty`, it isn't quite clear why I'm passing in an `instance=self`.

plankton/adapters/stream.py
```diff
@@ -103,11 +105,15 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
         to a string. The default map function only does that when the supplied value
         is not None.
 
+        Finally, documentation can be provided by passing the doc-argument. If it is omitted,
+        the docstring of the bound method is used and if that is not present, left empty.
```
MikeHart85, line 109: I think `__doc__` is always present, just set to `None` if no docstring has been given.

Might be a technicality, but could be worth rewording slightly to reflect that (and adjusting code accordingly).

plankton/adapters/stream.py
```diff
@@ -152,6 +159,24 @@ def __init__(self, device, arguments=None):
 
         self._create_properties(self.commands)
 
+    @property
+    def documentation(self):
+        cmd_template = '{}:\n{}'
```
MikeHart85: Nitpick: only used in one place on next line, arguably no need to have this as a separate variable.

plankton/adapters/stream.py
```diff
@@ -152,6 +159,24 @@ def __init__(self, device, arguments=None):
...
+
+        commands = [cmd_template.format(
+            cmd.pattern.pattern,
```
MikeHart85, line 166: This looks quite odd. 

I understand why it's like that, after having a look, but first impression is "what... :open_mouth:?"

plankton/core/utils.py
```diff
@@ -202,3 +203,20 @@ def failing_init(obj, *args, **kwargs):
...
+    manually added line breaks.
+
+    :param text: The text to format
+    :return: The formatted doc text
+    """
+
+    return '\n'.join(
+        textwrap.fill(line, width=99, initial_indent='    ', subsequent_indent='    ')
+        for line in
+        text.splitlines())
```
MikeHart85: Could this benefit from `inspect.cleandoc()` at all?

plankton/devices/chopper/interfaces/epics_interface.py
```diff
@@ -21,18 +21,65 @@
...
+        'ParkAng-RB': PV('target_parking_position', read_only=True,
+                         doc='Readback value of the discs parking position setpoint in degrees.'),
+        'ParkAng': PV('target_parking_position',
+                      doc='The discs parking position setpoint in degrees.'),
+        'AutoPark': PV('auto_park',
+                       doc='If enabled, the chopper disc will be moved to the parking '
+                           'position automatically when the speed is 0 or the chopper '
+                           'is otherwise stopped. 0 means False, 1 means True, the string '
+                           'representations of the enum values are "false" and "true".',
+                       type='enum', enums=['false', 'true']),
```
MikeHart85, line 81: Would it be better to place the documentation for all of these as docstrings on the actual device properties?

plankton/devices/linkam_t95/interfaces/stream_interface.py
```diff
@@ -21,6 +21,19 @@
...
+    Linkam T95 TCP stream interface
+
+    This is the interface of a simulated Linkam T95 device. The device listens on a configured
+    host:port-combination, one option to connect to it is via telnet:
+
+        $ telnet host port
+
+    Once connected, it's possible to send the specified commands, described in the dynamically
+    generated documentation. Information about host, port and line terminators in the concrete
+    device instance are also generated dynamically.
```
MikeHart85, line 34: As we talked about on Slack, `telnet` doesn't always work for this. Seems to depend on OS how well / if it works at all.

This has to do with the line endings. Linkam uses only a CR as a line terminator, while telnet will tend to use CRLF.

I think this is fine for this PR, but we should open an issue against this. The user needs a way to talk to the device on its primary protocol. 

We'll probably have to create our own little TCP Stream client that is aware of what line terminators the device its talking to expects.

MikeHart85:
Very valid question from @OwenArnold as he was glancing at this:

What happens when a device has two interfaces?

We should have at least one example device that has multiple interfaces to test for cases like that!

plankton/adapters/__init__.py
MichaelWedel: No, because a `ForwardProperty` sometimes wraps a plain member of the device instance (for example `chopper.speed`). In that case `getattr(type(getattr(...)))` fails, because it's not in the type and `getattr` on the instance would return `200` and the `__doc__` of `float`. I think that's okay if we want to continue allowing plain data members on devices. So instead of an error there will be no documentation if `doc` is not provided to `PV`.

plankton/adapters/__init__.py
MichaelWedel: You're right probably we don't need this here.

plankton/devices/chopper/interfaces/epics_interface.py
MichaelWedel: They are not always properties, but plain data members instantiated after the device is constructed. They can't have a docstring there. I only used `doc=...` where it wasn't otherwise possible.

plankton/adapters/__init__.py
MichaelWedel: The bright editor theme strikes again, thanks :) That should also clear up the comment below.

plankton/adapters/__init__.py
MikeHart85: In this case, consider using the `default` param of `getattr`:

``` python
self.__doc__ = getattr(type(getattr(instance, self._target_member)), self._prop, None).__doc__
```

In your scenario, this should effectively evaluate to `self.__doc__ = None.__doc__`, which in turn is equivalent to `self.__doc__ = None`, which is the default. No need for a try-except this way, and real exceptions don't get eaten.

plankton/devices/chopper/interfaces/epics_interface.py
MikeHart85: I see, didn't catch that :+1:

plankton/adapters/__init__.py
MichaelWedel: `instance` is required for actually retrieving the docstring, because the property can not be bound to a specific instance at this point. In principle it would be possible to take `type(instance)` instead, but it seemed easier to just ask for an object (which is `self` in the only place `ForwardProperty` is used).

plankton/adapters/__init__.py
MichaelWedel: That does look better. Didn't even realize until now that `getattr` has a default...

plankton/adapters/__init__.py
MichaelWedel: I looked at it and tried to remember a reason why I did that, but couldn't find any. It was probably a "carry-over" from `ForwardProperty` (where it also turned out to be unnecessary). In `StreamAdapter` this can never happen anyway, because an error is raised before `ForwardMethod` is constructed.

And you're right, it's no problem if this is an error already at this point because everything else in this will fail in that case...

plankton/adapters/stream.py
MichaelWedel: Hm, reading the sentence, it seems consistent to me...the docstring is absent (that may mean `__doc__` is `None`). How would you write it?

The code does not need any change, it's already doing `doc or __doc__ or ''`.

plankton/core/utils.py
MichaelWedel: I'm not too happy with `textwrap.fill`, it seems to be a problem wrapping text that already has linebreaks, which is the only reason for this splitting and re-combinding.

All the docstrings are obtained by `inspect.getdoc`. According to the documentation (if I remember correctly) that does apply `cleandoc` to the docstring. It could be added to the documentation that the function expects a "clean" docstring.

plankton/devices/chopper/interfaces/epics_interface.py
MichaelWedel: The case is a bit tricky...I don't think there is any way to attach a docstring to a plain data member (at least I couldn't come up with one).

plankton/devices/linkam_t95/interfaces/stream_interface.py
MichaelWedel: Yes, I agree. We had this open once I think...we could put it into `scripts` as well. For other protocols it may not be necessary to write one, so I don't think this will escalate to "a client for each protocol".

MichaelWedel:
Depends...multiple interfaces at the same time? For that we have to re-think more than this aspect anyway, because running the simulation is at the moment very much geared towards having one interface at a time. I think we're not there yet, but we should keep it in mind.

For multiple interfaces where only one is "active" it's a non-issue, because there's only one interface that can be queried for documentation.

plankton/adapters/stream.py
MichaelWedel: Yes..I wanted to avoid storing the pattern and the compiled regex, the name `pattern` is arguably a bit...unfortunate :)

plankton/devices/linkam_t95/interfaces/stream_interface.py
MichaelWedel: Ah yes, here is the gist:

https://gist.github.com/MichaelWedel/fd288ac19ef1d08d994e639be78aadd0

plankton/adapters/__init__.py
MikeHart85: :+1: I only found out while working on the StateMachine, came in handy in the same sort of situation where I wanted to avoid a `try-except` to not catch legitimate errors.

plankton/adapters/stream.py
MikeHart85: Looking over it, I can't see anything wrong with it either. I think I just wasn't reading it with text generation in mind, but just the value of `self.doc`. It's fine as-is.

plankton/core/utils.py
MikeHart85: Could `cleandoc`ing an already `cleandoc`ed string break anything? If not, it might be worthwhile doing it, since the function is a wrapper with the purpose of producing consistent output with inconsistent input. 

Feels odd to then have the documentation stating that your input must be consistent.

plankton/devices/chopper/interfaces/epics_interface.py
MikeHart85: Yeah, I don't think we should try anything like that. I just hadn't considered that case at all... was only thinking of methods and properties. :+1:

plankton/devices/linkam_t95/interfaces/stream_interface.py
MikeHart85: Very cool! I don't think I've seen that before (or I've just completely forgotten).

Yes, I think shipping something like that in scripts would be perfect.

If other protocols need it, we could extend it (or protocol developers could provide their own interface tool).

A potential alternative to consider would be to integrate both this sort of functionality and the current control client into a common entry point with subcommands (similar to docker or git):

```
$ plankton start chopper -r localhost:2020
$ plankton control localhost:2020 device
$ plankton connect localhost:9999
```

But we should probably create a separate issue for discussion on that.

PR #152, opened 2016-11-18
Create proper documentation using sphinx

### LINKED ISSUE #10: Add sphinx configuration to autogenerate documentation
I think we should automatically generate the documentation. @MikeHart85 already documented the `StateMachine`-class, it would make sense to use that.

MichaelWedel:
This fixes #10.

All markdown files have been transformed to rst files. A sphinx configuration file has been added to the `docs`-directory. I've chosen the RTD-theme because we are planning to deploy the documentation there after the renaming exercise anyway and this way it will look the same when users are creating it locally.

The documentation has a more clear structure now, with 3 "chapters":

- User guide
- Developer guide
- API Documentation

The latter is divided into the modules that we have. A lot of docstrings had to be added or moved slightly (this is one reason why the PR touches so many files, another is the rst-files that have been added under `docs`). I've tried to put as many meaningful cross-references in as possible so that it's easy to navigate the documentation (for example links to module/class/method doc-strings in the developer guide documents).

To build the documentation, follow the new instructions in the README.rst file.

Once the rename is done and the documentation is on RTD, we have to modify the readme and remove those specific links in favor of links to RTD. On the other hand, due to the sphinx-specific references and so on, some of the documents look pretty much broken on github now. Maybe the links should be removed, as it's pretty clear how to generate the documentation.

Finally, I have not added documentation for the specific devices to the docs yet, as we also have the dynamic documentation for these. I think it's better to treat them separately at a later point and try to get the dynamic aspect into the generated docs as well.

For review I think it's mainly generating the docs and going through them a bit to spot any major problems.


MikeHart85:
Getting a warning when building:
```
copying static files... WARNING: html_static_path entry u'/home/mikehart/Dev/plankton/docs/_static' does not exist
```
Googling around a bit, it seems like the way to avoid the warning is to do this in the conf.py:
```
- html_static_path = ['_static']
+ html_static_path = []
```

docs/index.rst
```diff
@@ -0,0 +1,29 @@
...
+   developer_guide/index
+   api/index
+
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
+* :ref:`search`
```
MikeHart85: This leads to a completely empty `search` page.

docs/index.rst
```diff
@@ -0,0 +1,29 @@
+Welcome to plankton's documentation!
+====================================
+
+Plankton is a Python package that makes it easy to develop complex stateful device simulations. It
+is licensed under GPL version 3 and the source is available on github_, where you are welcome to
+open new issues so the package can improve.
+
+If you're new to Plankton reading the :ref:`user_guide` is a good start. Once you plan on writing
+your own device it's time to go through the :ref:`developer_guide`. The :ref:`api_documentation`
```
MikeHart85: Not sure if this is worth doing anything about, but these generate links with anchors (e.g.: `api/index.html#api-documentation`) when really what is intended is just linking to the page. In the TOC below, the links appear without the anchor. This has the side-effect of fooling a browser into not showing both as visited unless the user has actually clicked both version, even though they link to the same thing.

One way to fix this would be to use `:doc:` instead of `:ref:`:

```
:doc:`developer_guide/index`
```
But that is a bit more verbose. So it's a question of which bugs us more...

docs/index.rst
```diff
@@ -0,0 +1,29 @@
...
+   user_guide/index
+   developer_guide/index
+   api/index
+
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
```
MikeHart85, line 26: Not sure if we can/need to do anything about this, but this leads to an alphabetized module index where every module is listed under one entry: `p - plankton`. Might be slightly better, if it's possible, to ignore the `plankton` and list them as `a - adapters`, `c - core`, etc?

docs/developer_guide/contributing.rst
```diff
@@ -0,0 +1,420 @@
+Contributing to Plankton
```
MikeHart85, line 1: Not sure if this PR is the place to do it, but I noticed this document has grown very long.

Might be good to split it up a bit, especially since it really shows when viewed in the RTD theme. On the other hand, it automatically generates subsections with anchor links... so maybe that's alright. It is noticeable that you have to scroll a lot more on this page than on others though.

Also, as we discussed before, this is more of a "How To Write Your Own Devices" document than a "Contributing" document. But that could be handled later, when we add real Contribution Guidelines.

docs/developer_guide/contributing.rst
```diff
@@ -0,0 +1,420 @@
...
+
+This entire device can be found in the ``examples`` directory. It can be
+started using the ``-k`` parameter of ``plankton.py``:
+
+::
+
+    $ ./plankton.py -k plankton.examples example_motor -- -b 127.0.0.1 -p 9999
+
+All functionality described in the
+:ref:`user_guide`, such as accessing the device and the simulation via the
```
MikeHart85, line 310: Same as previous comment, this could be ```:doc:`user_guide/index` ``` instead.

README.rst
```diff
@@ -0,0 +1,139 @@
...
+compatible with both Python 2 and 3.
+
+Plankton can be run directly using Python 2.7 or >= 3.4, or using a
+prepackaged Docker image that includes all dependencies. See relevant
+usage sections for details.
+
+Resources: -
+`GitHub <https://github.com/DMSC-Instrument-Data/plankton>`__ -
+`DockerHub <https://hub.docker.com/r/dmscid/plankton/>`__ -
+`Dockerfile <https://github.com/DMSC-Instrument-Data/plankton/blob/master/Dockerfile>`__
```
MikeHart85, line 16: This looks broken now... used to be a bullet point list. I actually like it better side-by-side, but then the first `-` should be removed. And potentially use a `--` or `|` as the separator between links?

README.rst
```diff
@@ -0,0 +1,139 @@
...
+   door"
+-  Access and control the ``Simulation`` while it is running via a "back
+   door"
+
+Additional Documentation
+------------------------
+
+Details on running Plankton, working with Plankton as a device
+developer, and framework internals are described in the following
+documents:
```
MikeHart85, line 88: The list following this used to be nested and now looks quite odd with everything at one level.

But I agree with you that ideally, the whole list should be removed and replaced with a link to online documentation.

docs/index.rst
MichaelWedel: On the one hand it would also save the labels, but on the other hands probably labels are not the worst thing here in case we change the structure of this. I'm leaning a bit more towards the `:doc:`-solution you suggested.

MikeHart85:
Since this contains documentation generated from code, will it be auto-generated/updated on merging into master? Only manually when releasing? 

Would it be possible to have/keep versioned documentation for all releases + `edge` development? I'm thinking something along the lines of Qt documentation, where the version is in the URL.

docs/index.rst
MichaelWedel: I'll see what I can find out about this, this was actually just the default document.

docs/index.rst
MichaelWedel: Maybe we don't need this. Or maybe it might work online? We could also add it later on...

docs/developer_guide/contributing.rst
MichaelWedel: I think I agree with renaming the document, but I also think that everything in that document belongs together...and especially with the anchored subsections it's still okay to navigate.

README.rst
MichaelWedel: Ah I didn't even notice that. Shouldn't be a problem to fix.

README.rst
MichaelWedel: This should also be unproblematic to fix.

docs/index.rst
MikeHart85: I was a bit surprised to find that the search box at the top of the panel on the left works offline. Not sure what sort of Voodoo is involved there, I'm a bit scared to look.

I kind of suspect that empty search page is just a template for results to be put in? Just speculating though.

MichaelWedel:
I'm still confused as to why I can reply to "review" comments but not to "comment" comments.

When documentation is generated can be configured in RTD. You can have "latest" for the latest master and "release" versions for a specific release. But I have not looked at this in detail yet, that would be a separate issue I think.

docs/developer_guide/contributing.rst
MikeHart85: Sounds good to me! :+1:

MikeHart85:
I know, it's weird and annoying you can't reply to general comments!

Sounds good on looking into it later, was just a thought I had as I was looking over this. :+1:

docs/index.rst
MichaelWedel: I added `modindex_common_prefix = ['plankton.']` to the configuration, that does exactly what you suggested.

docs/index.rst
MichaelWedel: Didn't find anything in particular that was helpful, so I'm just removing the link from that page. There is a search-box in the docs anyway.

docs/index.rst
MichaelWedel: I decided to link to the docs directly, but keep the labels for linking to this from elsewhere (in case the structure ever changes).

docs/developer_guide/contributing.rst
MichaelWedel: Here I think it's actually not so bad to use `:ref:`, because that makes linking more flexible (i.e. the document name doesn't need to be known).

MikeHart85:
Looks good, but the formatting for the nest listed was still a bit off (nested items were appearing as a quote). I fiddled a bit with it and it should look alright now. Apparently in RST, the nested bullet points need to be aligned with the first text character of the parent node, AND there must be a blank line between the first child and the parent node... very picky.

Went ahead and committed the change, can you check it looks good to you?

Everything else looks good to me and ready to merge... but codacy is complaining about `copyright` being redefined now. Do you know what that is about? It seemed happy with it in your initial push, and it hasn't changed since.

Issue #151, opened 2016-11-11
Error displayed when Device not found


MikeHart85:
If you try to start a device that doesn't exist or cannot be found, e.g.:

```
$ ./plankton.py foobar
```

You will get an error like this:

```
An error occurred:
Could not find setup 'default' for device 'foobar'.
Are you sure the correct device package is used (flag -k) and that it is in the path (flag -a)?
```

This can be confusing, especially when the device you're trying to load has no setup. It should really say that it couldn't find the device, and not mention setups.

MichaelWedel:
I'm tempted to say we can fix it as part of #132, where all of these things are going to change anyway.

MichaelWedel:
This was resolved in PR #154, the error is now:
```
$ ./plankton.py foobar
An error occurred:
No device with the name 'foobar' could be found. Possible names are:
    chopper
    linkam_t95
See also the -k option to add inspect a different module.
```

PR #154, opened 2016-11-26
Introduce device & interface factory

### LINKED ISSUE #132: Introduce factory for device and interface
Currently device and interface discovery and construction are scattered over a few functions in a few different modules.

The plan is to move this to a factory which collects this in one place. The factory should also have tests.

MichaelWedel:
This fixes #132 and #151.

This consolidates `import_device` and `import_adapter` in one central place. As a consequence, some of the functions related to this in `core.utils` changed slightly or became obsolete.

For manual testing please check that the existing devices still work (i.e. running the example_motor and the linkam), using the `-k` flag. Please also take a look at the documentation (the new module is in core/devices).

Having two interfaces with the same protocol name will result in an error quite early (see unit tests). I've also spent some time testing setups in multiple locations. My conclusion is that if there is a `setups` package with individual modules inside, this will be used and the `setups` dict is apparently "overwritten" at some point.

All error messages regarding device discovery have been worked on. For example:

```
$ ./plankton.py -k plankton.examples foobar
An error occurred:
No device with the name 'foobar' could be found. Possible names are:
    simple_device
    example_motor
See also the -k option to add inspect a different module.
```
Similar messages have been introduced for setups and protocols.


plankton/adapters/__init__.py
```diff
@@ -85,78 +83,6 @@ def handle(self, cycle_delay=0.1):
...
 
 
-def is_adapter(obj):
-    try:
-        return issubclass(obj, Adapter) and not obj.__module__.startswith('plankton.adapters')
-    except TypeError:
-        return False
-
-
-def get_available_adapters(device_name, device_package):
```
MikeHart85, line 95: I think it's great these are getting removed from `adapters/__init__.py`, always felt out of place here. 

Do you think it would be appropriate to move the remaining classes (`Adapter`, `ForwardProperty` and `ForwardMethod`) to `core/adapter.py` or something similar? Not necessarily in this PR, just a general question. I'm always surprised to find real code in `__init__.py` files, usually I expect them to be empty or just contain some namespacing glue.

plankton/core/utils.py
```diff
@@ -25,39 +25,78 @@
...
+from os import path as osp
 from os import listdir
 
+from datetime import datetime
+
 from .exceptions import PlanktonException
 
 
-def get_available_submodules(package):
+def get_submodules(module):
```
MikeHart85, line 40: These functions (`get_submodules`, `get_members`, `extract_module_name`) are only ever used in `core/devices.py`.

Would it be more appropriate to move them there? They don't seem like functions that would be used in any other context or like general purpose utility functions.

plankton/devices/__init__.py
```diff
@@ -187,95 +186,3 @@ def _override_data(self, overrides):
...
 
                 setattr(self, name, val)
-
-
-def is_device(obj):
-    return issubclass(obj, CanProcess) and obj.__module__ not in (
-        'plankton.core.processor', 'plankton.devices')
-
-
-def import_device(device, setup=None, device_package='devices'):
```
MikeHart85, line 197: Similar to a previous comment, it's great to see these removed from here.

Could the remaining `Device` and `StateMachineDevice` classes be moved somewhere else? Could be added to `core/devices.py`. Alternatively, could move them to a `core/device.py` together with `DeviceBase` and rename the remainder of `core/devices.py` to something more general (since it also handles finding interfaces and setups)... `discovery.py` or `registry.py`?

Again, no change necessary for this PR, just a general comment.

plankton/scripts/run.py
```diff
@@ -73,41 +71,34 @@ def do_run_simulation(argument_list=None):
...
+        print('\n'.join(device_builder.protocols))
         return
 
-    device = device_type(**parameters)
-    adapter = import_adapter(
-        arguments.device, arguments.protocol,
-        device_package=arguments.device_package)(device, arguments.adapter_args)
+    device = device_builder.create_device(arguments.setup)
+    interface = device_builder.create_interface_type(arguments.protocol)(
+        device, arguments.adapter_args)
```
MikeHart85: It looks like this is the only real (non-test) usage of `create_interface_type`. Might be appropriate to refactor it to just accept all three arguments and return an instance directly?

plankton/core/devices.py
```diff
@@ -0,0 +1,349 @@
...
+        try:
+            return self._create_device_instance(
+                device_type, **setup_data.get('parameters', {}))
+        except RuntimeError:
+            raise PlanktonException(
+                'The setup \'{}\' you tried to load does not specify a valid device type, but the '
+                'device module \'{}\' provides multiple device types so that no meaningful '
+                'default can be deduced.'.format(setup_name, self.name))
+
+    def create_interface_type(self, protocol=None):
```
MikeHart85: As referenced in another comment, couldn't this be changed to:

```python
def create_interface(self, protocol, device, args):
    ...
```
I can't find any usage of the default protocol value (None is explicitly passed in when not specified on command line) or needing to get the type rather than an instance.

plankton/adapters/__init__.py
MichaelWedel: Yes, I've started going in that direction in the Adapter vs. Interface change that we've been discussing. I agree that there should be nothing in `__init__.py` for this module.

plankton/core/utils.py
MichaelWedel: You're right in that they are not used elsewhere, but they are fairly general. I wouldn't want to put them into `devices` mainly because of their very general nature...maybe we can find a better place than `utils` at some point, but for now I'd rather leave them there.

plankton/devices/__init__.py
MichaelWedel: I agree with this as well.

plankton/core/utils.py
MikeHart85: Alright, sounds good! :+1:

plankton/scripts/run.py
MichaelWedel: What I kind of don't like with this approach is that it drags in this entire adapter command lines arguments thing into the factory. I'd be more comfortable with a `**kwargs` type solution to avoid duplicating the adapter signature here. I'll try and implement that.

MikeHart85: Sounds good. I'd also include `*args` to support calling it without being verbose about argument names.

PR #157, opened 2016-12-08
Add check_limits decorator to core.utils

MichaelWedel:
This has no issue number, but it came up while experimenting with the socket device.

A new decorator is introduced called `check_limits`, which can check the parameter passed to a unary function. The main intention was use on property setters:

```
class Foo(object):
    _bar = 5

    @property
    def bar(self):
        return self._bar

    @bar.setter
    @check_limits(-4.0, 6.7)
    def bar(self, new_bar):
        self._bar = new_bar
```

For more flexibility regarding updating limits at runtime either of the limits can be a string, which is then assumed to be a plain data member of the object:

```
class Foo(object):
    _bar = 5
    _bar_min = -4.0

    @property
    def bar(self):
        return self._bar

    @bar.setter
    @check_limits('_bar_min', 6.7)
    def bar(self, new_bar):
        self._bar = new_bar
```

You can also suppress the exception by supplying `silent=True` to the decorator. This would be useful for the interface of the linkam (could also profit from `argument_mappings`):

```
    # Cmd('set_limit', ..., argument_mappings=(int,), return_mapping=lambda x: "" if x is None else str(x))
    @check_limits(-2000, 6000, silent=True):
    def set_limit(self, limit):
        self._device.temperature_limit = limit / 10.0
```

### DIFF
--- a/plankton/core/exceptions.py
+++ b/plankton/core/exceptions.py
@@ -29,3 +29,10 @@ class PlanktonException(Exception):
     from unexpected ones. This enables better error handling and more importantly
     better presentation of errors to the users.
     """
+
+
+class LimitViolationException(Exception):
+    """
+    An exception that can be raised in a device to indicate a limit violation. It is for example
+    raised by the :class:`~plankton.core.utils.check_limits`.
+    """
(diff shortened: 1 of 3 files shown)


MikeHart85:
As discussed, should open some issues to generalize this more, but it's good to go as is to start with.

PR #163, opened 2016-12-16
Update documentation to reflect new installation method

### LINKED ISSUE #162: Update documentation to new distribution method
As pointed out by @MikeHart85, now that there's a package on PyPI we should update the documentation to reflect that. Since it also installs scripts that make it easier to use, the "Usage with Python" section needs to be updated. This is also a good opportunity to promote the use of virtual environments.

MichaelWedel:
Now that the package is available on PyPI, the user guide was updated to reflect that change. I also added a very brief section on virtual environments, with a link to the documentation of the `virtualenv` package for further details.

Please build the documentation locally and check that the pages look okay and that the text makes sense.

This closes #162.

### DIFF
--- a/docs/user_guide/remote_access_devices.rst
+++ b/docs/user_guide/remote_access_devices.rst
@@ -10,7 +10,7 @@ the ``-r`` option with a ``host:port`` string to the simulation:
 
 ::
 
-    $ python lewis.py -r 127.0.0.1:10000 chopper -- -p SIM:
+    $ lewis -r 127.0.0.1:10000 chopper -- -p SIM:
 
 Now the device can be controlled via the ``lewis-control.py``-script
 in a different terminal window. The service can be queried to show the
@@ -18,7 +18,7 @@ available objects by not supplying an object name:
 
 ::
 
-    $ python lewis-control.py -r 127.0.0.1:10000
+    $ lewis-control -r 127.0.0.1:10000
 
 The ``-r`` (or ``--rpc-host``) option defaults to the value shown here,
 so it will be omitted in the following examples. To get information on
@@ -27,7 +27,7 @@ method will list the object's API:
 
 ::
 
-    $ python lewis-control.py device
+    $ lewis-control device
 
 This will output a list of properties and methods which is available for
 remote access. This may not comprise the full interface of the object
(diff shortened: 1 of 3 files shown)


docs/user_guide/remote_access_devices.rst
```diff
@@ -53,7 +53,7 @@ command line:
     $ python lewis-control.py device start
 
 Only numeric types and strings can be used as arguments via the
-``lewis-control.py``-script. The script always tries to convert
+``lewis-control``-script. The script always tries to convert
 parameters to ``int`` first, then to ``float`` and leaves it as ``str``
 if both fail. For other types and more control over types, it's advised
```
MikeHart85: I don't think this statement is accurate anymore. We now use `ast.literal_eval`, which should support any kind of literal.

docs/user_guide/usage_with_python.rst
```diff
@@ -64,3 +150,6 @@ directory):
 
 Details about parameters for the various adapters, and differences
 between OSes are covered in the "Adapter Specifics" sections.
+
+If you decided to install Lewis this way, please be aware that the ``lewis`` and ``lewis-control``
+calls in the other parts of the guide have to be replaced with ``python lewis.py``.
```
MikeHart85, line 155: You could also just execute them directly (`$ path/to/lewis.py ...`) since they have a shebang and +x permissions. 

Not sure if running it that way works on Windows though? If it doesn't, it's probably not worth changing anything here. `python lewis.py` will definitely work cross-platform.

docs/user_guide/remote_access_devices.rst
MichaelWedel: Very good point, I'll change that.

docs/user_guide/usage_with_python.rst
MichaelWedel: Not sure about Windows either. The entire thing is quite *nix-centric with the shell and so on, so I don't know how well all of this works on Windows.

MikeHart85: Let's just leave this as-is then. `python lewis.py` is pretty universal.

MikeHart85:
I was just going to make a tiny edit on one line, but I somehow ended up adding a couple of sections to this page. I think we briefly talked about expanding a bit on how to set things like dicts and lists using the control utility at some point, but never got around to it. 

Felt like we might as well do it here since we're touching up that section anyway. Have a look to see whether you like the changes. We could also move them to a separate PR if you think that would be better.

I'm ready to merge this, just waiting for your review of my changes.

MichaelWedel:
Cool, thanks for updating the section, it is much clearer now. The examples really add a lot, especially for lists and dicts the syntax is a bit less straightforward than it could be. From my point of view it's good to go.

PR #165, opened 2016-12-16
Added Julabo waterbath emulation

mattclarke:
I have add an emulator for a Julabo waterbath.
The Julabo protocol is slightly/annoyingly different across different models, so I have created two protocol files based on two of the Julabos we have at ISIS.
There will be others...

Hopefully, this is up to standard.


mattclarke:
Didn't notice that the name change had taken place. Have rebased my changes.

MikeHart85:
This looks great Matt!

If you have any complaints/ideas/issues you ran into while making this, please let us know. It would be great to get some feedback on the practicalities of writing emulators in lewis from outside our team.

lewis/devices/julabo/device.py
```diff
@@ -0,0 +1,302 @@
...
+            (('circulate', 'not_circulate'), lambda: not self.circulate_commanded),
+        ])
+
+    def get_bath_temperature(self):
+        """
+        Gets the external temperature of the bath.
+
+        :return: The external temperature.
+        """
+        return self.temperature
```
MikeHart85: I've noticed there are a lot of trivial getters. 

The `PV` helper class of the EPICS adapter allows referencing attributes directly to expose them without a need for explicit getters or properties (e.g.: `PV('target_speed', read_only=True, ...)` exposes `self.target_speed = 0.0` from `SimulatedChopper` as a read-only PV). 

`Cmd` currently only supports exposing methods. I think this was simply a consequence of the fact that the Linkam T95 being the first device we looked at for TCP Stream, and it having no getters except for T Status... which wasn't trivial at all and made sense as a method.

Would support for exposing attributes directly with `Cmd` have made writing this emulator easier / cleaner? Do you think it would make sense for stream devices in general? Would perhaps a second helper class called `Var` make sense, to allow trivial reading and/or writing to attributes?

MichaelWedel: I really like the idea of having `Var`. It could have a `read_cmd` and `write_cmd` (both regular expressions) and would automatically be readable, writeable or both depending on which are specified.

I'm going to open an issue, but I think we can merge this in with no problems first, the `Var` change can be done as a refactoring step.

lewis/devices/julabo/device.py
```diff
@@ -0,0 +1,302 @@
...
+    def get_status(self):
+        """
+        Not sure what a real device returns as the manual is a bit vague.
+        It will return error codes but it is not clear what it returns if everything is okay.
+
+        :return: String
+        """
+        return "Hello from the simulated Julabo"
+
+    def set_set_point(self, param):
```
MichaelWedel, line 74: This would be an example for `check_limits` (very new), which would simplify the function to:

```
@check_limits('temperature_low_limit', 'temperature_high_limit', silent=True)
def set_set_point(self, new_setpoint):
    self.set_point_temperature = new_setpoint
    return ""
```

The `return ""` part could even be achieved by having a `return_mapping` on the device's interface. Would it help to make it possible to have a default return mapping on the interface? Then it would only need to be overridden for cases where it's different. Here it's probably something like:

```
def default_return_mapping(value):
    return "" if value is None else value
```

lewis/devices/julabo/interfaces/julabo_stream_interface_2.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    protocol = "julabo-version-2"
+
+    commands = {
+        Cmd('get_bath_temperature', '^IN_PV_00$'),
+        Cmd('get_external_temperature', '^IN_PV_01$'),
+        Cmd('get_power', '^IN_PV_02$'),
+        Cmd('get_set_point', '^IN_SP_00$'),
+        Cmd('get_high_limit', '^IN_SP_03$'),    # Different from version 1
```
MichaelWedel: Seeing the comment I'm wondering whether it could be an option to allow for inheritance here and then overriding bits of the `commands` member. On the other hand it would mean that one has to be more careful when changing things in the base protocol.

MikeHart85: Is there anything preventing inheritance right now? Haven't tried, but as long as you're careful to append to `commands` rather than overwriting it... it might work? Not overrides, but you could factor out the common functionality into a base class and append version specifics in child classes?

MichaelWedel: No, nothing is preventing that, but it would look different compared to other interface definitions.

mattclarke: Hi both.

Happy New Year!

Sorry for not commenting earlier, but Xmas hols got in the way.

I like what you have done with Var - I think it will make life easier. I'll update my code when I get a chance.

The whole process of writing an emulator was pretty straightforward; I needed some help from MikeHart85 with the different protocols but other than that it was a case of following the examples. You have done a great job 🥇

mattclarke: Hi both.

Happy New Year!

Sorry for not commenting earlier, but Xmas hols got in the way.

I like what you have done with Var - I think it will make life easier. I'll update my code when I get a chance.

The whole process of writing an emulator was pretty straightforward; I needed some help from MikeHart85 with the different protocols but other than that it was a case of following the examples. You have done a great job 🥇

mattclarke: I think the Julabo is pretty unique in having slightly different protocols across models, so unless there are other use cases you can think of I wouldn't worry too much about it.

MichaelWedel: Thanks for your feedback and especially for that use case - we didn't think of something like `Var` before. One thing that we also discovered in our discussions is for the case of trivial getter and non-trivial setter, where you can define both `Var` and `Cmd`, for example like this:

```python
from lewis.adapters.stream import StreamAdapter, Cmd, Var
from lewis.core.utils import check_limits

class Interface(StreamAdapter):
    commands = {
        Var('param', read_pattern='^P$', doc='Some param')
        Cmd('set_param', pattern='^P=([0-9]+)$, argument_mappings=(int,))
    }

    param = 10

    @check_limits(0, 20)
    def set_param(self, new_param):
        """Set param after checking limits."""
        self.param = new_param
```

Then you get a trivial getter and a limit checked, type converted setter.

MikeHart85: Thanks Matt!

Would you prefer we go ahead and merge this as-is for now, or keep it open for you to work on still?

There's no rush to merge it on our end, so it down to which way is more convenient for you.

mattclarke: I'll make the changes first and let you know when they are done. It shouldn't take long.

mattclarke: Var is really cool - it has reduced the amount of code significantly.

As a suggestion it would be really nice is if you could pass Var limits to check then I could get rid of the setters that do nothing other than check the limits, like:

```python
    @check_limits(0, 999)
    def set_external_d(self, param):
        """
        Sets the external derivative.
        Tv in Julabo speak.

        :param param: The value to set, must be an integer between 0 and 999
        :return: Empty string.
        """
        self.external_d = param
        return ""
```

It might not be much use generally, so feel free to disregard.

lewis/devices/julabo/interfaces/julabo_stream_interface_2.py
```diff
@@ -29,28 +29,32 @@ class JulaboStreamInterfaceV2(StreamAdapter):
...
         Cmd('set_internal_d', '^OUT_PAR_08 ([0-9]*)$', argument_mappings=(int,)),
-        Cmd('get_external_p', '^IN_PAR_09$'),
-        Cmd('get_external_i', '^IN_PAR_11$'),
-        Cmd('get_external_d', '^IN_PAR_12$'),
+        Var('external_p', read_pattern='^IN_PAR_09$', doc='The external proportional.'),
         Cmd('set_external_p', '^OUT_PAR_09 ([0-9]*\.?[0-9]+)$', argument_mappings=(float,)),
+        Var('external_i', read_pattern='^IN_PAR_11$', doc='The external integral.'),
         Cmd('set_external_i', '^OUT_PAR_11 ([0-9]*)$', argument_mappings=(int,)),
+        Var('external_d', read_pattern='^IN_PAR_12$', doc='The external derivative.'),
         Cmd('set_external_d', '^OUT_PAR_12 ([0-9]*)$', argument_mappings=(int,)),
```
MichaelWedel, line 58: We're working on a way to make `check_limits` a bit more flexible so that it would be possible to use that here, for example like this:

```python
class JulaboStreamInterfaceV2(StreamAdapter):
    commands = {
        Var('external_d', read_pattern='^IN_PAR_12$',
            write_pattern='^OUT_PAR_12 ([0-9]*)$', argument_mappings=(check_limits(0, 999)(int),))
    }
```
Then the only thing the setters are still doing is returning an empty string instead of `None`. We have briefly discussed a solution for this as well. It can be done now with `return_mapping`, but we might rename it and make it possible to introduce a default in the class, it could be something like this:

```python
class JulaboStreamInterfaceV2(StreamAdapter):
    default_return_policy = lambda x: str(x) if x is not None else ""
```
If the behavior is different in a particular `Cmd`, it can be overridden there with `return_policy` (we might rename `return_mapping`). Right now each `Var`/`Cmd` would have to explicitly specify this.

Sorry for the many modifications, all this helps to make this sort of thing simpler for the future (hopefully).

lewis/devices/julabo/device.py
```diff
@@ -20,10 +20,29 @@
...
+    heating_power = 5.0  # The heating power
+    version = "JULABO FP50_MH Simulator, ISIS"
+    status = "Hello from the simulated Julabo"
+    is_circulating = 0  # 0 for off, 1 for on
+    temperature = 24.0  # Current temperature in C
+    external_temperature = 26.0  # External temperature in C
+    circulate_commanded = False
+    temperature_ramp_rate = 5.0  # Guessed value in C/min
+
     def _initialize_data(self):
```
MichaelWedel, line 46: This is completely empty now and can be removed. As lewis is evolving it might not even be necessary at all in the (very near) future, so the way you defined the members as part of the class is fine. Everything that lewis allows you to do now (namely overriding those values in a setup) is compatible with the way you wrote it here.

MichaelWedel (commented):
Thanks for the work you're putting into this Matt, with your help we're able to iron out a few more inconsistencies and problems in the framework and as written in the other comment, the process of adding a new device should be much less painful next time :)

lewis/devices/julabo/interfaces/julabo_stream_interface_2.py
MichaelWedel: I realized that `check_limits(0, 999)(int)` is not doing the right thing, it's checking limits first and then converting to integer. A proper solution will take a bit more time, so for now I suggest leaving it as it is.

mattclarke: Okay, will do.

PR #167, opened 2016-12-18
Add Var to complement Cmd

### LINKED ISSUE #166: Introduce Var in addition to Cmd
As pointed out in #165, with the current implementation of `Cmd` and `StreamAdapter`, many trivial getters and/or setters have to be added potentially.

To solve this problem, an additional class `Var` should be introduced which allows for easier exposure of properties/attributes. The way `StreamAdapter` works at the moment, it heavily depends on the internals of `Cmd`, this should be changed so that `StreamAdapter` only accesses the interface of `Cmd`, which `Var` should share. Instead of `Var` inheriting from `Cmd` it might be useful to have something like `CmdBase` as a common base of the two which defines the interface.

Having thought about this a bit more, given some more time and a few more iterations, this could also promote further harmonisation with `PV`.

MichaelWedel:
This fixes #166.

By doing a bit more work than just introducing `Var`, `Cmd` is more flexible now and the entire process of binding the commands to the device are better defined and separated now.

A person implementing a device interface still uses `Cmd` in exactly the same way as before, minus the `regex_flags` parameter. In addition, it's now also possible to specify a callable object as the first parameter instead of a member name, which can be a free function or a lambda. Both `Cmd` and `Var` have a `bind`-method, which returns one or more `Func`-objects. Those contain all the logic that is required for determining whether a request can be processed, argument mapping, return value mapping and actually calling the associated function.

Now it's no longer necessary to modify the type of the interface to forward device methods/properties, this is done in the binding step instead.

Another side effect is that it should be massively easier to unit test this stuff, because it's all encapsulated now.

### DIFF
--- a/lewis/adapters/stream.py
+++ b/lewis/adapters/stream.py
@@ -184,8 +184,9 @@ class Cmd(BaseCommand):
 
 
 class Var(BaseCommand):
-    def __init__(self, target_member, read_pattern=None, write_pattern=None, argument_mappings=None,
-                 return_mapping=lambda x: None if x is None else str(x), doc=None):
+    def __init__(self, target_member, read_pattern=None, write_pattern=None,
+                 argument_mappings=None, return_mapping=lambda x: None if x is None else str(x),
+                 doc=None):
         self.member = target_member
 
         self._patterns = {key: re.compile(pattern) for key, pattern in
--- a/lewis/adapters/stream.py
+++ b/lewis/adapters/stream.py
@@ -21,14 +21,14 @@ from __future__ import print_function
 
 import asynchat
 import asyncore
+import inspect
 import re
 import socket
-import inspect
 from argparse import ArgumentParser
 
 from six import b
 
-from lewis.adapters import Adapter, ForwardMethod
+from lewis.adapters import Adapter, ForwardMethod, ForwardProperty
 from lewis.core.utils import format_doc_text
 
 
@@ -50,14 +50,12 @@ class StreamHandler(asynchat.async_chat):
         try:
             match = None
             for cmd in self.target.commands:
-                match = cmd.pattern.match(request)
-                if match:
-                    groups = match.groups()
-                    func = getattr(self.target, cmd.method)
-
-                    args = cmd.map_arguments(groups)
-                    reply = cmd.return_mapping(func(*args))
+                try:
+                    reply = cmd.process_request(request, self.target)
+                    match = True
                     break
+                except CanNotProcessException:
+                    pass
 
             if match is None:
                 raise RuntimeError('None of the device\'s commands matched.')
@@ -86,7 +84,44 @@ class StreamServer(asyncore.dispatcher):
             StreamHandler(sock, self.target)
 
 
-class Cmd(object):
+class CanNotProcessException(Exception):
+    pass
+
+
+class BaseCommand(object):
+    member = None
+    argument_mappings = None
+    return_mapping = None
+    doc = None
+
+    def process_request(self, request, target):
+        raise NotImplementedError('Commands must implement process_request method.')
+
+    @property
+    def patterns(self):
+        raise NotImplementedError('Commands must implement patterns property.')
+
+    def map_arguments(self, arguments):
+        """
+        Returns the mapped function arguments. If no mapping functions are defined, the arguments
+        are returned as they were supplied.
+
+        :param arguments: List of arguments for bound function as strings.
+        :return: Mapped arguments.
+        """
+        if self.argument_mappings is None:
+            return arguments
+
+        return [f(a) for f, a in zip(self.argument_mappings, arguments)]
+
+    def map_return_value(self, return_value):
+        if self.return_mapping is None:
+            return return_value
+
+        return self.return_mapping(return_value)
+
+
+class Cmd(BaseCommand):
     """
     This is a small helper class that makes it easy to define commands that are parsed
     by StreamAdapter and forwarded to the correct methods on the Adapter.
@@ -108,16 +143,15 @@ class Cmd(object):
 
     :param target_method: Method to be called when regex matches.
     :param regex: Regex to match for method call.
-    :param regex_flags: Flags to pass ot re.compile, default is 0.
     :param argument_mappings: Iterable with mapping functions from string to some type.
     :param return_mapping: Mapping function for return value of method.
     :param doc: Description of the command. If not supplied, the docstring is used.
     """
 
-    def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
+    def __init__(self, target_method, regex, argument_mappings=None,
                  return_mapping=lambda x: None if x is None else str(x), doc=None):
-        self.method = target_method
-        self.pattern = re.compile(b(regex), regex_flags)
+        self.member = target_method
+        self.pattern = re.compile(b(regex))
 
         if argument_mappings is not None and (self.pattern.groups != len(argument_mappings)):
             raise RuntimeError(
@@ -128,18 +162,71 @@ class Cmd(object):
         self.return_mapping = return_mapping
         self.doc = doc
 
-    def map_arguments(self, arguments):
-        """
-        Returns the mapped function arguments. If no mapping functions are defined, the arguments
-        are returned as they were supplied.
+    @property
+    def patterns(self):
+        return [self.pattern.pattern]
 
-        :param arguments: List of arguments for bound function as strings.
-        :return: Mapped arguments.
-        """
-        if not self.argument_mappings:
-            return arguments
+    def process_request(self, request, target):
+        match = self.pattern.match(request)
 
-        return [f(a) for f, a in zip(self.argument_mappings, arguments)]
+        if not match:
+            raise CanNotProcessException()
+
+        args = self.map_arguments(match.groups())
+        func = getattr(target, self.member)
+
+        return self.map_return_value(func(*args))
+
+
+class Var(BaseCommand):
+    def __init__(self, target_member, read_regex=None, write_regex=None, argument_mappings=None,
+                 return_mapping=lambda x: None if x is None else str(x), doc=None):
+        self.member = target_member
+
+        self.read_pattern = re.compile(read_regex) if read_regex is not None else None
+        if self.read_pattern.groups != 0:
+            raise RuntimeError(
+                'Command regex for reading member \'{}\' contains '
+                'arguments.'.format(target_member))
+
+        self.write_pattern = re.compile(write_regex) if write_regex is not None else None
+
+        if self.write_pattern.groups != 1:
+            raise RuntimeError(
+                'Command regex for writing member \'{}\' contains more '
+                'than one argument.'.format(target_member))
+
+        self.argument_mappings = argument_mappings
+        self.return_mapping = return_mapping
+        self.doc = doc
+
+    @property
+    def patterns(self):
+        patterns = []
+
+        if self.read_pattern is not None:
+            patterns.append(self.read_pattern)
+
+        if self.write_pattern is not None:
+            patterns.append(self.write_pattern)
+
+        return patterns
+
+    def process_request(self, request, target):
+        if self.read_pattern is not None:
+            match = self.read_pattern.match(request)
+
+            if match:
+                return self.map_return_value(getattr(target, self.member))
+
+        if self.write_pattern is not None:
+            match = self.write_pattern.match(request)
+
+            if match:
+                args = self.map_arguments(match.groups())
+                return self.map_return_value(setattr(target, self.member, *args))
+
+        raise CanNotProcessException()
 
 
 class StreamAdapter(Adapter):
@@ -241,20 +328,26 @@ class StreamAdapter(Adapter):
     def _create_properties(self, cmds):
         patterns = set()
         for cmd in cmds:
-            method = cmd.method
+            member = cmd.member
 
-            if method not in dir(self):
-                if method not in dir(self._device):
-                    raise AttributeError('Can not find method \''
-                                         + method + '\' in device or interface.')
-                setattr(self, method, ForwardMethod(self._device, method))
+            if member not in dir(self):
+                if member not in dir(self._device):
+                    raise AttributeError('Can not find member \''
+                                         + member + '\' in device or interface.')
 
-            if cmd.pattern.pattern in patterns:
+                if callable(getattr(self._device, member)):
+                    setattr(self, member, ForwardMethod(self._device, member))
+                else:
+                    setattr(type(self), member, ForwardProperty('_device', member, instance=self))
+
+            cmd_patterns = set(cmd.patterns)
+
+            if not cmd_patterns.isdisjoint(patterns):
                 raise RuntimeError(
                     'The regular expression \'{}\' is '
-                    'associated with multiple methods.'.format(cmd.pattern.pattern))
+                    'associated with multiple members.'.format(cmd.pattern.pattern))
 
-            patterns.add(cmd.pattern.pattern)
+            patterns.update(cmd_patterns)
 
         if len(patterns) < len(cmds):
             raise RuntimeError('Warning')
--- a/lewis/adapters/stream.py
+++ b/lewis/adapters/stream.py
@@ -115,10 +115,13 @@ class BaseCommand(object):
         return [f(a) for f, a in zip(self.argument_mappings, arguments)]
 
     def map_return_value(self, return_value):
-        if self.return_mapping is None:
-            return return_value
+        if callable(self.return_mapping):
+            return self.return_mapping(return_value)
 
-        return self.return_mapping(return_value)
+        if self.return_mapping is not None:
+            return self.return_mapping
+
+        return return_value
 
 
 class Cmd(BaseCommand):
@@ -136,7 +139,9 @@ class Cmd(BaseCommand):
 
     The return_mapping argument is similar, it should map the return value of the method
     to a string. The default map function only does that when the supplied value
-    is not None.
+    is not None. It can also be set to a numeric value or a string constant so that the
+    command always returns the same value. If it is ``None``, the return value is not
+    modified at all.
 
     Finally, documentation can be provided by passing the doc-argument. If it is omitted,
     the docstring of the bound method is used and if that is not present, left empty.
@@ -148,10 +153,10 @@ class Cmd(BaseCommand):
     :param doc: Description of the command. If not supplied, the docstring is used.
     """
 
-    def __init__(self, target_method, regex, argument_mappings=None,
+    def __init__(self, target_method, pattern, argument_mappings=None,
                  return_mapping=lambda x: None if x is None else str(x), doc=None):
         self.member = target_method
-        self.pattern = re.compile(b(regex))
+        self.pattern = re.compile(b(pattern))
 
         if argument_mappings is not None and (self.pattern.groups != len(argument_mappings)):
             raise RuntimeError(
@@ -179,22 +184,24 @@ class Cmd(BaseCommand):
 
 
 class Var(BaseCommand):
-    def __init__(self, target_member, read_regex=None, write_regex=None, argument_mappings=None,
+    def __init__(self, target_member, read_pattern=None, write_pattern=None, argument_mappings=None,
                  return_mapping=lambda x: None if x is None else str(x), doc=None):
         self.member = target_member
 
-        self.read_pattern = re.compile(read_regex) if read_regex is not None else None
-        if self.read_pattern.groups != 0:
+        self._patterns = {key: re.compile(pattern) for key, pattern in
+                          zip(('read', 'write'), (read_pattern, write_pattern)) if
+                          pattern is not None}
+
+        if 'read' in self._patterns and self._patterns['read'].groups != 0:
             raise RuntimeError(
                 'Command regex for reading member \'{}\' contains '
                 'arguments.'.format(target_member))
 
-        self.write_pattern = re.compile(write_regex) if write_regex is not None else None
-
-        if self.write_pattern.groups != 1:
+        if 'write' in self._patterns and self._patterns['write'].groups != 1:
             raise RuntimeError(
-                'Command regex for writing member \'{}\' contains more '
-                'than one argument.'.format(target_member))
+                'Command regex for writing member \'{}\' is expected to contain one '
+                'argument, but it contains {}.'.format(
+                    target_member, self._patterns['write'].groups))
 
         self.argument_mappings = argument_mappings
         self.return_mapping = return_mapping
@@ -202,25 +209,17 @@ class Var(BaseCommand):
 
     @property
     def patterns(self):
-        patterns = []
-
-        if self.read_pattern is not None:
-            patterns.append(self.read_pattern)
-
-        if self.write_pattern is not None:
-            patterns.append(self.write_pattern)
-
-        return patterns
+        return [pattern.pattern for pattern in self._patterns.values()]
 
     def process_request(self, request, target):
-        if self.read_pattern is not None:
-            match = self.read_pattern.match(request)
+        if 'read' in self._patterns:
+            match = self._patterns['read'].match(request)
 
             if match:
                 return self.map_return_value(getattr(target, self.member))
 
-        if self.write_pattern is not None:
-            match = self.write_pattern.match(request)
+        if 'write' in self._patterns:
+            match = self._patterns['write'].match(request)
 
             if match:
                 args = self.map_arguments(match.groups())
--- a/lewis/adapters/stream.py
+++ b/lewis/adapters/stream.py
@@ -24,7 +24,6 @@ import asyncore
 import inspect
 import re
 import socket
-from itertools import chain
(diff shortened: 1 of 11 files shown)


MichaelWedel:
Sorry, had forgotten some print statements, force-pushed the branch without them.

MichaelWedel:
Codacy is lying - I'm pretty sure that under the circumstances it's being used as a callable, it's callable. Unless of course someone specifies something that is not callable on purpose.

lewis/adapters/stream.py
```diff
@@ -86,7 +84,44 @@ def handle_accept(self):
...
 
-class Cmd(object):
+class CanNotProcessException(Exception):
+    pass
+
+
+class BaseCommand(object):
+    member = None
+    argument_mappings = None
+    return_mapping = None
```
MikeHart85: It might be too late to change this, but I don't like the term `mapping` here very much. It's not wrong, but a bit too general to be useful... kind of like naming a class `SomethingManager`. Something along the lines of `arg_types` and `return_filter` might be more descriptive, and more intuitive to understand what's happening when a user sees things like: 

```python
Cmd('set_target', r'^T=([-+]?[0-9]*\.?[0-9]+)$', arg_types=(float,)),
Cmd('stop', r'^H$', return_filter=lambda ret: 'T={},P={}'.format(ret[0], ret[1])),
```

But changing it now would break code that already uses these names. Do you think it would be worthwhile?

lewis/adapters/stream.py
```diff
@@ -86,7 +84,44 @@ def handle_accept(self):
             StreamHandler(sock, self.target)
 
 
-class Cmd(object):
+class CanNotProcessException(Exception):
+    pass
```
MikeHart85: This sort of goes back to #135... I can see how it is convenient to make ad hoc exception types where you need them to catch only that specific exception type, but is this good practice in general?

lewis/adapters/stream.py
```diff
@@ -50,14 +50,12 @@ def found_terminator(self):
...
-                    func = getattr(self.target, cmd.method)
-
-                    args = cmd.map_arguments(groups)
-                    reply = cmd.return_mapping(func(*args))
+                try:
+                    reply = cmd.process_request(request, self.target)
+                    match = True
                     break
+                except CanNotProcessException:
+                    pass
```
MikeHart85: Consider returning a tuple and using instead a calling convention like this:

```python
match, reply = cmd.process_request(request, self.target)
```

This would avoid the need for a custom exception and a try-except block here.

lewis/adapters/stream.py
```diff
@@ -108,16 +143,15 @@ class Cmd(object):
 
     :param target_method: Method to be called when regex matches.
     :param regex: Regex to match for method call.
-    :param regex_flags: Flags to pass ot re.compile, default is 0.
```
MikeHart85, line 111: Since we're removing the `regex_flags` argument, we should briefly mention something about the `(?iLmsux)` feature and how to use it in the documentation above.

lewis/adapters/stream.py
```diff
@@ -128,18 +162,71 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
...
-        :return: Mapped arguments.
-        """
-        if not self.argument_mappings:
-            return arguments
+    def process_request(self, request, target):
+        match = self.pattern.match(request)
 
-        return [f(a) for f, a in zip(self.argument_mappings, arguments)]
+        if not match:
+            raise CanNotProcessException()
```
MikeHart85: As mentioned above, this could be replaced with a `return False, None`, and the successful return below with a `return True, ...`

lewis/adapters/stream.py
```diff
@@ -128,18 +162,71 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
...
+class Var(BaseCommand):
+    def __init__(self, target_member, read_regex=None, write_regex=None, argument_mappings=None,
+                 return_mapping=lambda x: None if x is None else str(x), doc=None):
+        self.member = target_member
+
+        self.read_pattern = re.compile(read_regex) if read_regex is not None else None
+        if self.read_pattern.groups != 0:
+            raise RuntimeError(
+                'Command regex for reading member \'{}\' contains '
+                'arguments.'.format(target_member))
```
MikeHart85: While unlikely, there is a possibility that someone might want to use a capturing group not as an argument, but to backreference within the regular expression. For example, `r'([a-z]+)[0-9]+\1'` will match `'foo123foo'` and `'bar123bar'`, but not `'foo123bar'`. Treating this as an error prevents that usage.

But the same limitation exists in `Cmd`, since all capturing groups are treated as arguments.

Not something to worry about here, just to consider.

lewis/adapters/stream.py
```diff
@@ -128,18 +162,71 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
...
+    def patterns(self):
+        patterns = []
+
+        if self.read_pattern is not None:
+            patterns.append(self.read_pattern)
+
+        if self.write_pattern is not None:
+            patterns.append(self.write_pattern)
+
+        return patterns
```
MikeHart85: For brevity, this function could be reduced to:

```python
def patterns(self):
    return filter(lambda x: x is not None, [self.read_pattern, self.write_pattern])
```

Or even (if we don't care that this will also remove `0`, `False` and `''`... and we shouldn't?):

```python
def patterns(self):
    return filter(None, [self.read_pattern, self.write_pattern])
```

lewis/examples/simple_device/__init__.py
```diff
@@ -39,8 +39,7 @@ class VerySimpleInterface(StreamAdapter):
     After that, typing either of the commands and pressing enter sends them to the server.
     """
     commands = {
-        Cmd('get_param', '^P$'),
-        Cmd('set_param', '^P=(.+)$'),
+        Var('param', read_regex='^P$', write_regex='^P=(.+)$')
```
MikeHart85: Maybe keep the `Cmd` version and make the regex here match `V` instead (while still exposing `param`) just to illustrate both are possible.

lewis/adapters/stream.py
```diff
@@ -128,18 +162,71 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
...
+
+            if match:
+                return self.map_return_value(getattr(target, self.member))
+
+        if self.write_pattern is not None:
+            match = self.write_pattern.match(request)
+
+            if match:
+                args = self.map_arguments(match.groups())
+                return self.map_return_value(setattr(target, self.member, *args))
```
MikeHart85: This could be a problem. Setters may need to return any number of things. Sometimes they return the new or old value. Sometimes they always return an empty string. Sometimes they don't return anything. 

You can do any of that here... but you don't want it affecting your getter.

lewis/adapters/stream.py
```diff
@@ -241,20 +328,26 @@ def _parseArguments(self, arguments):
...
+
+            if member not in dir(self):
+                if member not in dir(self._device):
+                    raise AttributeError('Can not find member \''
+                                         + member + '\' in device or interface.')
+
+                if callable(getattr(self._device, member)):
+                    setattr(self, member, ForwardMethod(self._device, member))
+                else:
+                    setattr(type(self), member, ForwardProperty('_device', member, instance=self))
```
MikeHart85: Is the `type(self)` rather than `self` intended?

lewis/adapters/stream.py
```diff
@@ -86,7 +84,44 @@ def handle_accept(self):
...
+        if self.argument_mappings is None:
+            return arguments
+
+        return [f(a) for f, a in zip(self.argument_mappings, arguments)]
+
+    def map_return_value(self, return_value):
+        if self.return_mapping is None:
+            return return_value
+
+        return self.return_mapping(return_value)
```
MikeHart85: I don't know if this will make Codacy happy (to be fair, it has a point), but you could do this instead:
```python
    def map_return_value(self, return_value):
        if callable(self.return_mapping):
            return self.return_mapping(return_value)

        return return_value
```
Or even:
```python
    def map_return_value(self, return_value):
        if callable(self.return_mapping):
            return self.return_mapping(return_value)
        if self.return_mapping is not None:
            return self.return_mapping

        return return_value
```
This would have the side-effect of allowing unconditional, static return overrides in addition to lambda's and other callables.

MikeHart85 (commented):
You still have it tagged `in progress`, but I had some feedback/comments as you're working on it.

MichaelWedel:
That's what the `In Progress` was intended for - discussions, because it's not finished at all.

lewis/adapters/stream.py
MichaelWedel: It goes a bit with the comment above, I tried with different return values first, but somehow returning a mix of error code (what match essentially is) and `reply`, which is the actual result didn't really appeal to me - especially since I'm not aware we're using this "pattern" elsewhere.

On the other hand this is kind of "misusing exceptions for flow control", but only because there is no other use of `process_request` (at the moment). Otherwise I think raising an exception would be the right thing to do here - "I tried processing the request, I wasn't able to do it, maybe someone else is". I also used a special type, because that is an exception that is expected and we know how to deal with it. Other exceptions may result from actually executing the function(s) and for those we actually do not know and we catch them very broadly to forward them to the `handle_error` method.

A completely different way would be to have an additional method `can_process(request)` which just checks the regex and returns True or False. In case of actually processing the command, the request would then be run against the regex again to extract the groups. Or the `process_request` method takes a regex match object, but somehow I don't think having those in such a general method signature is particularly nice either...

I first thought this could be some sort of "Chain of Responsibility" pattern, but I think that would require a new type that is returned from the function that makes it possible to distinguish between "this is the original request" and "this is the result of the processing".

Any other ideas how to handle this?

lewis/adapters/stream.py
MichaelWedel: I'm not sure if `return_filter` would tell me much more than `return_mapping`. To me a filter is something that removes something from something else, which is a valid use case for this, but not the only one. Maybe `transformation` was what I had in mind as a use case.

Similar for `argument_mappings` - while most of the time it might be type transformations, it could be other things as well, transforming to lower case, taking a substring, ...

Would `return_transformation` or `argument_transformations` be better? In general I think it would be okay to change it, but only if we find something that is substantially better and captures all possible uses for these two.

lewis/adapters/stream.py
MichaelWedel: I like that a lot, it looks much better. But it doesn't make the naming easier :)

lewis/adapters/stream.py
MichaelWedel: Agreed.

lewis/adapters/stream.py
MichaelWedel: Yes, I agree that we should keep that in mind for another round of improvement - in this change I'd say we have enough other things to worry about even if we don't change the behavior of `Cmd`.

lewis/adapters/stream.py
MichaelWedel: In the light of `filter`, `map` and `reduce` being almost removed from Python 3 at some point a list comprehension would probably be "future proof" (tm).
```
def patterns(self):
    return [x for x in (self.read_pattern, self.write_pattern) if x is not None]
```

No matter which one, it would indeed make the function shorter.

lewis/adapters/stream.py
MichaelWedel: Yes, it's required for installing a property.

lewis/examples/simple_device/__init__.py
MichaelWedel: Good idea, will make the modifications.

lewis/adapters/stream.py
MichaelWedel: What do you propose? Having a getter mapping and a setter mapping? That would certainly solve the issue, but the question is do we want to do that?

lewis/adapters/stream.py
MikeHart85: I had completely forgotten that `argument_mappings` could be arbitrary transformation functions.

Not sure I have better terms. In a way, these are pre- and post-processing... filters for lack of a better word.. Probably best to leave it as-is in that case.

MikeHart85: `_override`, `_transform`, `_mutate`, `_process`, `_filter` (I wouldn't say this necessarily removes; consider audio/video/image filters)

MichaelWedel: True, did not consider video filters and so on (that was probably the chemist speaking). I kind of like the term `transform`, because that's what it actually is right?

`argument_transforms`, `return_transform`?

lewis/adapters/stream.py
MikeHart85: I'd never heard of `filter`, etc being deprecated or slated for removal. Searching for it led me to some really old (2003-2005) discussions revolving around a comment PEP 279. My feeling is that if it hasn't happened 13 years later, it's pretty safe to assume there is too much opposition to removing them and that they will stay.

I like `filter` here, especially the `None` variant. It's concise and clear, much more so than the list comprehension, which requires more... comprehending to understand. My vote is to use `filter` and prepare the pitchforks and torches in case some faceless tyrant tries to take it away from us.

lewis/adapters/stream.py
MikeHart85: I'm not sure, I hadn't gotten that far. I just flagged it as a potential problem in the back of my mind back when I suggested `Var` in #165 (I pictured it exactly the way you implemented it here... with optional read and write regexes and all), so I knew to look for this here.

Is there anything preventing a user from splitting into two `Var`s when needed? So doing this:

```python
Var('param', read_regex='^P$', return_mapping=...),
Var('param', write_regex='^P=(.+)$', return_mapping=...),
```
Instead of a single `Var` for both? Only for the case you need custom return mappings?

If you can do this, it might not be a problem as long as it is documented.

MikeHart85: Ah, on that note... also need to be careful about calling the arguments `read_regex` and `write_regex` (and `regex` in `Cmd`).

We might want to implement `scanf` formatting as we discussed at some point. With `Cmd` it wasn't so critical because that argument wouldn't typically be named. But with `Var`, we would likely always name the argument when using it.

Should probably call them something very generic like just `read` and `write`... or maybe `read_format`, etc.

lewis/adapters/stream.py
MikeHart85: Interesting, I'll have to dig a bit to understand what's going on there.

lewis/adapters/stream.py
MichaelWedel: This is how different views can be for a simple thing...for me personally the list comprehension is easier to understand, because I haven't used `filter` in years and don't know what the it does if the first argument is `None`.

lewis/adapters/stream.py
MichaelWedel: There's nothing preventing the user from splitting it up into two `Var`s, so that is probably the easiest solution and we should document that in terms of some example.

That's what I meant on Slack that I wasn't happy with the parameter naming either. Maybe `read_pattern`, `write_pattern` and `pattern` are better, they wouldn't conflict with either `scanf` or `regex`.

MichaelWedel: Ah there, `read_format`, that could work as well, although I would maybe find it slightly unexpected in the context of regular expressions.

lewis/adapters/stream.py
MikeHart85: Transform makes me think of matrices and 3D graphics. 

The old fashioned coder in me also yearns for something much shorter... I'd probably call them `arg_map` and `ret_map`, or maybe `arg_filter` and `ret_filter`, if I was writing this. I'd leave the `s` implied in the common understanding that there can be multiple arguments but only one return value... but then, that's the C coder talking again and we're in Python land.

But this is down to personal preferences now... so I'll leave it entirely up to you. I'm happy we discussed these names a bit, but we shouldn't dwell too much on it.

MichaelWedel: Well in this case I'd say we take the easy way out and leave it as it is until somebody tells us "I have no idea what you're talking about".

lewis/adapters/stream.py
MikeHart85: Well, I'll admit it is a bit misleading... but in a good(?) way here!

It does exactly this with a `None` parameter:
```[x for x in (self.read_pattern, self.write_pattern) if x]```
(three `x`s, ugh...)

Note the missing `is not None` at the end. It simply checks for truthiness. We only care about Noneness, but it wouldn't hurt to filter the others here.

File: lewis/adapters/stream.py (hunk 9 above)
Comment (MikeHart85) [reply lewis/adapters/stream.py]: `*_pattern` sound great to me!

File: lewis/adapters/stream.py (hunk 1 above)
Comment (MikeHart85) [reply lewis/adapters/stream.py]: Sounds like a plan :smile:

File: lewis/adapters/stream.py (hunk 10 above)
Comment (MichaelWedel) [reply lewis/adapters/stream.py]: This is the whole data descriptor thing...on the object it looks like an attribute, so you can never tell from the object whether it's that or a `property`. I'm not sure my understanding is good enough to explain it in a proper way...

But I do think `ForwardProperty` could be implemented in a different way, more like in `control_client`.

File: lewis/adapters/stream.py (hunk 7 above)
Comment (MichaelWedel) [reply lewis/adapters/stream.py]: While working on it I found some more problems (this needs to be `x.pattern` for one) and the logic in `__init__` was wrong. I'll push in a minute.

File: lewis/adapters/stream.py [hunk 12]
```diff
@@ -128,18 +167,66 @@ def __init__(self, target_method, regex, regex_flags=0, argument_mappings=None,
...
+    def __init__(self, target_member, read_pattern=None, write_pattern=None,
+                 argument_mappings=None, return_mapping=lambda x: None if x is None else str(x),
+                 doc=None):
+        self.member = target_member
+
+        self._patterns = {key: re.compile(pattern) for key, pattern in
+                          zip(('read', 'write'), (read_pattern, write_pattern)) if
+                          pattern is not None}
+
+        if 'read' in self._patterns and self._patterns['read'].groups != 0:
```
MichaelWedel: The first part of the check wasn't there, so it would have failed for cases with only a read or write pattern. Collecting these in a dict solves the problem of `patterns` as well, no filter required there anymore, because read/write pattern are not in the dict.

lewis/examples/simple_device/__init__.py
```diff
@@ -31,16 +31,25 @@ class VerySimpleInterface(StreamAdapter):
...
+     - ``V``: Returns the parameter as part of a verbose message.
+     - ``V=something``: Sets the parameter to ``something``.
+     - ``P``: Returns the device parameter unmodified.
+     - ``P=something``: Exactly the same as ``V=something``.
+
     """
     commands = {
-        Cmd('get_param', '^P$'),
-        Cmd('set_param', '^P=(.+)$'),
+        Cmd('get_param', pattern='^V$', return_mapping=lambda x: 'The value is {}'.format(x)),
```
MichaelWedel: This now demonstrates the use of `return_mapping`.

MichaelWedel:
Codacy is still not happy, although the check for callable is right there. The duplication issue is somewhat valid and these simple assignments could go into the base class `__init__`.

lewis/adapters/stream.py
MikeHart85: Thought about this some more. It's a bit awkward no matter how you do it. But I'm still leaning towards the tuple return.

Introducing `can_process` would give us a way to check before calling `process_request`, but then you have to use the same match in both places. And I could just call `process_request` with garbage values anyway... so it would still need to throw an exception on no match. So, in effect, we haven't simplified anything and only added code duplication and complexity.

Chain of responsibility seems like overkill for this simple situation. We could consider a simplified version, like what the `re` module does: Take parameters normally but return a `match` object that may or may not have results, and contains meta information about the nature and validity of results.

But what would our return object need to look like / what values would it contain? It would contain only two things: 

1. A boolean to indicate whether the request was accepted/processed, and 
2. A return value from processing the request (default `None` if above `False`). 

So: `(False, None)` or `(True, ret_value)`. Why create a new type for this if tuple fits the bill?

MikeHart85:
I would treat the codacy issue about non-callable as a bug on their end (or PyLint I suppose) at this point.

At most, we could replace `return_mapping = None` initialization with something callable to help it understand... but that's just silly and would be wrong.

lewis/adapters/stream.py
MichaelWedel: From my point of view "And I could just call process_request with garbage values anyway... so it would still need to throw an exception on no match." is the key argument here. This is how I would write the function if someone told me "Write a function that processes something and return the result". I would make it raise an exception, I'm just not entirely sure which one...

lewis/adapters/stream.py
```diff
@@ -48,20 +48,11 @@ def found_terminator(self):
...
-                if match:
-                    groups = match.groups()
-                    func = getattr(self.target, cmd.method)
-
-                    args = cmd.map_arguments(groups)
-                    reply = cmd.return_mapping(func(*args))
-                    break
-
-            if match is None:
+            try:
```
MichaelWedel: The only thing left here that I do not like is the nested `try...except`. Otherwise `next` seems like a good solution to this problem. I don't really care that we perform the "winning match" twice, it's a small cost to pay for avoiding all the other stuff.

`process_request` still raises an exception now, but a standard one, so we do not need a new type here.

lewis/adapters/stream.py
```diff
@@ -48,20 +48,13 @@ def found_terminator(self):
...
-                    func = getattr(self.target, cmd.method)
-
-                    args = cmd.map_arguments(groups)
-                    reply = cmd.return_mapping(func(*args))
-                    break
-
-            if match is None:
+            cmd = next((cmd for cmd in self.target.commands if cmd.can_process(request)), None)
+
+            if cmd is None:
```
MichaelWedel, line 53: Now this is much better from my point of view. Get's the first matching `Cmd` or `None` if none is found. In that case, as before, raise `RuntimeError` so it can be handled (or not). Otherwise continue processing request.

What do you think?

lewis/examples/simple_device/__init__.py
```diff
@@ -31,16 +31,27 @@ class VerySimpleInterface(StreamAdapter):
     A very simple device with TCP-stream interface
 
     The device has only one parameter, which can be set to an arbitrary
-    value. The interface consists of two commands which can be invoked via telnet.
+    value. The interface consists of four commands which can be invoked via telnet.
```
MikeHart85: Surely you mean `five`. :wink:

lewis/examples/simple_device/__init__.py
```diff
@@ -31,16 +31,27 @@ class VerySimpleInterface(StreamAdapter):
...
+     - ``R`` or ``r``: Returns the number 4.
+
     """
     commands = {
-        Cmd('get_param', '^P$'),
-        Cmd('set_param', '^P=(.+)$'),
+        Cmd('get_param', pattern='^V$', return_mapping='The value is {}'.format),
+        Cmd('set_param', pattern='^V=(.+)$', argument_mappings=(int,)),
+        Var('param', read_pattern='^P$', write_pattern='^P=(.+)$', doc='The only parameter.'),
+        Cmd(lambda: 4, pattern='^R$(?i)', doc='Random number')
```
MikeHart85: The `Random` might warrant being `"Random"` instead.

lewis/adapters/stream.py
```diff
@@ -86,47 +79,83 @@ def handle_accept(self):
...
+    def __init__(self, func, pattern, argument_mappings=None, return_mapping=None, doc=None):
+        if not callable(func):
+            raise RuntimeError('Can not construct a Func-object from a non callable object.')
+
+        self.func = func
+        self.raw_pattern = pattern
+        self.pattern = re.compile(b(pattern), 0) if pattern else None
+
+        func_args = num_args(func)
+        if self.pattern.groups != func_args:
```
MikeHart85: You might be better off using `inspect.getcallargs` ([link](https://docs.python.org/2/library/inspect.html#inspect.getcallargs)) for this purpose. That function will allow you to *"try"* calling `func` without actually `try`ing to call it, if that makes sense.

Consider:
```python
inspect.getcallargs(func, *[None]*self.pattern.groups)
```
This will return some stuff (that you don't care about here) *IF it is possible* to call `func` with that number of arguments (without actually calling it). If it is NOT possible to call it with that amount of arguments, it will raise a `TypeError` exception.

This way, you don't need `num_args` at all from what I can tell (and the tests for it can go away as well). A shortcoming of this `num_args` implementation is that it doesn't consider optional arguments with default values, or `*args` / `**kwargs` scenarios... if `func` has anything like that, it will reject valid regexes / group counts.

MichaelWedel: Thanks for that, I was considering `try...except TypeError...` but with actually calling the function, which isn't really safe.

I checked `inspect` but most likely did not read closely enough.

MikeHart85: It is sort of abusing the function... this isn't exactly what it was meant for, but it just fits so nicely.

The only reason I thought of it is because I used `getcallargs` for something else recently. It's a very useful function in general.

PR #169, opened 2016-12-23
EPICS meta data can be updated at runtime

### LINKED ISSUE #158: Runtime adjustable EPICS metadata
Currently it's only possible to set PV control metadata such as unit, limits, etc. once in `PV`, for example:

```
class Foo(EpicsAdapter):
    pvs = {
        'pv': PV('pv', unit='mm')
    }
```

For something like `unit` this type of static behaviour might be okay in most cases, but sometimes it needs to be updated at runtime, for example for limits. `pcaspy` has a method `setParamInfo`, which allows for updating this type of meta information.

The `PV` object should have an optional argument `metadata_property` (or similar), which points to a property similar to the first argument of the constructor. It should specify a property which may return a meta data update based on internal device state. This should be consumed by `PropertyExposingDriver` along with the general PV update procedure.

To then actually check those limits and reject `caput` operations which would exceed them, pre-requisite work has been done as part of PR #157.

MichaelWedel:
This fixes #158.

With this change it is now possible to supply PV with an additional argument `metadata_property` that returns a dict which updates PV metadata. Additionally, LimitViolationExceptions are now caught in PropertyExposingDriver so that the `check_limits` decorator can be used to enforce PV limits.

### DIFF
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -27,7 +27,7 @@ from . import Adapter, ForwardProperty
 from six import iteritems
 
 from lewis.core.utils import seconds_since, FromOptionalDependency, format_doc_text
-from lewis.core.exceptions import LewisException
+from lewis.core.exceptions import LewisException, LimitViolationException
 
 # pcaspy might not be available. To make EPICS-based adapters show up
 # in the listed adapters anyway dummy types are created in this case
@@ -57,16 +57,40 @@ class PV(object):
     are forwarded to the pcaspy server's `pvdb`, so it's possible to pass on
     limits, types, enum-values and so on.
 
+    In case those arguments change at runtime, it's possible to provide ``meta_data_property``,
+    which should contain the name of a property that returns a dict containing these values.
+    For example if limits change:
+
+    .. sourcecode:: Python
+
+        class Interface(EpicsAdapter):
+            pvs = {
+                'example': PV('example', meta_data_property='example_meta')
+            }
+
+            @property
+            def example_meta(self):
+                return {
+                    'lolim': self._device._example_low_limit,
+                    'hilim': self._device._example_high_limit,
+                }
+
+    The PV infos are then updated together with the value, determined by ``poll_interval``.
+
     :param target_property: Property of the adapter to expose.
     :param poll_interval: Update interval of the PV.
     :param read_only: Should be True if the PV is read only.
+    :param meta_data_property: Property which returns a dict containing PV metadata.
     :param doc: Description of the PV. If not supplied, docstring of mapped property is used.
     :param kwargs: Arguments forwarded into pcaspy pvdb-dict.
     """
-    def __init__(self, target_property, poll_interval=1.0, read_only=False, doc=None, **kwargs):
+
+    def __init__(self, target_property, poll_interval=1.0, read_only=False,
+                 meta_data_property=None, doc=None, **kwargs):
         self.property = target_property
         self.read_only = read_only
         self.poll_interval = poll_interval
+        self.meta_data_property = meta_data_property
         self.doc = doc
         self.config = kwargs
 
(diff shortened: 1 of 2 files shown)


MichaelWedel:
Does this need any further changes?

MikeHart85:
Having trouble seeing how this fits in with everything else... some questions:

1. Do we have any real use case for this?
1. How does/will this tie in with `@check_limits`, if at all? What about distinguishing `HIGH` / `LOW` vs `HIHI` / `LOLO`?
1. Where would metadata in general come from? Since `Device`s must be protocol-agnostic, they must not know anything about EPICS-specific metadata (or its values) like `unit`, `type`, etc. So it must be coded in the `Interface`, right? What would this look like in practice?
1. Have you considered direct setting over callback update? E.g.:
```python
class Foo(EpicsAdapter):
    pvs = {
        'foobar': PV('foobar', unit='mm'),
        'unit': PV('unit', type='string')
    }

    @property
    def unit(self):
        return self.pvs['foobar'].getParamInfo('unit')

    @property.setter
    def unit(self, value):
        self.pvs['foobar'].setParamInfo('unit', value)
```
Does this callback approach have advantages over something like the above?

MichaelWedel:
The use case for this is a device like this:

```python
class SomeDevice(StateMachineDevice):
    p_hilim = 10.0
    p_lolim = 0.0
    _p = 1.0

    @property
    def p(self):
        return self._p

    @p.setter
    @check_limits('p_lolim', 'p_hilim')
    def p(self, new_p):
        self._p = new_p
```
In reality someone might go to the device and push the button to change the limit. Or log into the device and change it. In a simulation it would probably happen through the control server.

The corresponding Interface:

```python
class SomeDeviceEpicsInterface(EpicsAdapter):
    pvs = {
        'P' : PV('p', meta_data_property='p_meta', unit='mm')
    }

    @property
    def p_meta(self):
        return {
           'lolim': self._device.p_lolim,
           'hilim': self._device.p_hilim,
        }
```

These are not separate PVs, this is the information that is contained in the response from for example `epics.PV.get_ctrlvars` from `pyepics`. Each PV has them, although they might not always be defined by all IOCs.

I did not see a better way to make these kind of changes at runtime possible and it ties in with `check_limits` as well, as these are the "control limits". The other limits defined by EPICS are as far as I understand warning levels.

Does this answer your questions?

MikeHart85:
I see, that makes much more sense thanks!

One alternative could be to make a `MetaPV` object (which effectively is just a reference to an attribute by name), which would look like this:
```python
class SomeDeviceEpicsInterface(EpicsAdapter):
    pvs = {
        'P' : PV('p', unit='mm', lolim=MetaPV('p_lolim'), hilim=MetaPV('p_hilim'))
    }
```
This would remove the need for defining a `p_meta` that returns a dict in the Interface. The `MetaPV` instances could also do some smart things like caching the old value so you can update only when it changes. 

Not sure if there would be any value to that. But it would likely be much more work in the Adapter to glue all this together. I like how little code your solution needs to implement this, this is just an idea I had while looking over it.

lewis/adapters/epics.py
```diff
@@ -101,6 +125,10 @@ def process_pv_updates(self, force=False):
             if self._timers[pv] >= pv_object.poll_interval or force:
                 try:
                     self.setParam(pv, getattr(self._target, pv_object.property))
+
+                    if pv_object.meta_data_property is not None:
+                        self.setParamInfo(pv, getattr(self._target, pv_object.meta_data_property))
```
MikeHart85, line 130: Unlike `setParam`, `setParamInfo` calls `updatePVs` every time it is called. Since it's called in a loop here, that means we could get multiple `updatePVs` calls in a row in a single cycle.

I'm not sure if this is an issue in any way... just something to keep an eye on.

MikeHart85:
Looks good to merge to me, let me know if still want to make changes.

lewis/adapters/epics.py
MichaelWedel: Thanks, I was not aware of this! I'm not sure it's an issue either, I agree on keeping an eye on it and seeing if it becomes a problem at some point.

MichaelWedel:
I think for this iteration I don't want to change anything else. You're definitely right about the caching aspect, I totally agree on that. But in principle that argument applies to the PV value as well, so we should probably go back to this method and `PV` in general to see if we can make it better. It would be something for the next release, this time it was `StreamAdapter` that got all the attention.

Issue #144, opened 2016-11-03
Adapter traffic being queued while disconnected


MikeHart85:
This is a bug reported by @IsabellaRey during an IBEX meeting.

Commands/requests sent to the device while it is disconnected are queued, then processed when it is reconnected. To reproduce:

1. Launch a `linkam_t95` simulator in plankton, with control server
1. Use control client to disconnect device
1. `telnet` into `linkam_t95` simulator, type `T` and press enter
1. Use control client to connect device

You will get a TArray reply after the last step.

This is a result of the fact that we don't actually disconnect but just stop processing the adapter. TCP/IP stack apparently still queues requests and gives them to us once we start processing again.

One possible solution is flushing all requests when reconnecting.

Side Issue: Step 3 arguably shouldn't be possible / doesn't make sense. Again, a consequence of not really disconnecting. Will make another issue for this.

MichaelWedel:
I did not expect this to work this way, I'm particularly unsure why step 3 works. We really do have to change this.

MikeHart85:
Yes, this is definitely unexpected and not really representative of any real-world scenario I can think of.

The reason step 3 works is because we simply stop calling `asyncore.loop()` when a device is "disconnected", but the `StreamServer(asyncore.dispatcher)` is still up (by design) and apparently it doesn't need loop cycles to receive `handle_accept` calls (I'm surprised by this).

Basically, I think there are two different scenarios we should simulate in this context:
1. Disconnect: The device is physically disconnected (or fails completely). In this case, all connected clients should be disconnected immediately and no further connections should be accepted. Attempts to connect should be rejected.
2. Suspend: The device stops responding/hangs/lags (potentially temporarily). In this case, current connections should stay open, but the device should stop responding. Attempts to connect should time out.

I'm not sure what should happen with commands sent in scenario 2 though (this issue, essentially). Should they be queued, as they are, or lost completely?

MichaelWedel:
The bug as described above has been fixed with #170, it is no longer possible to connect while the device is disconnected and connections are properly cut, so no queuing anymore. The suspend/resume issue is now in #171 - can we close this?

MikeHart85:
Well... really, #170 hasn't fixed this so much as moved it from "disconnected" to "suspended" and temporarily removed the suspend feature. Should probably leave this open until #171 implements suspending in a way that fixes this. Or we could close this one, but then at least make sure #171 is clear that it's taking over this issue and is responsible for fixing it.

[...]

PR #170, opened 2017-01-02
Fix disconnecting devices

### LINKED ISSUE #145: Could we implement real disconnecting as opposed to suspending?
Currently,  `disconnect_device` and `connect_device` functionality of the Simulation is implemented by suspending Adapter processing. The Adapter no longer gets cycles and so the device appears to stop responding.

Some issues with this are highlighted in #144.

Also, conceptually, this isn't really the same as disconnecting. 

The reason we took this approach is because of concerns about just closing and freeing up the port that the device simulator is listening on. If we simply free the port, other applications or simulators may claim it.

This issue is to explore two questions:

1. Would it be technically possible to better simulate a disconnect scenario without risking losing the port? For example, could all clients be disconnected and further connections refused?
1. The current behaviour could still be useful to simulate scenarios were a device stops responding (without being disconnected). But "disconnect" isn't really the right term for this. Should we rename current functionality to `suspend_device` / `revive_device` or similar?

MichaelWedel:
This fixes #145.

`Adapter` now has a method `stop_server` in addition to `start_server`, as well as a `is_running` property which indicates the status of the adapter. Both in the EPICS and Stream cases, the entire server object is completely torn down in `stop_server`, so that existing connections are closed and new connections can not be made. As discussed in #144 and/or #145, there is a risk that the port is blocked by another process in the meantime, but currently all of that is the user's responsibility anyway so at the moment I do not see a big problem with this. Should we ever pursue the "simulators as a service" route, we have to re-think guarantees in that respect anyway.

The implementation of `Simulation` has been changed to make use of the new `Adapter` capabilities.

For testing the `simple_example` could be used:

Start a stream based device:
```
$ ./lewis.py -r localhost:10000 -k lewis.examples simple_device
```

Connect to it, make sure you get something from the device:
```
$ telnet localhost 9999
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
P
10
```

Disconnect the device:
```
./lewis-control.py simulation disconnect_device
```

The telnet session should close. For EPICS a `camonitor` should show the same behavior.

The suspend/resume problem has been transferred to a new issue #171.

### DIFF
--- a/lewis/adapters/__init__.py
+++ b/lewis/adapters/__init__.py
@@ -63,7 +63,7 @@ class Adapter(object):
 
     def start_server(self):
         """
-        This method should be re-implemented to start the infrastructure required for the
+        This method must be re-implemented to start the infrastructure required for the
         protocol in question. These startup operations are not supposed to be carried out on
         construction of the adapter in order to preserve control over when services are
         started during a run of a simulation.
@@ -75,11 +75,13 @@ class Adapter(object):
 
         .. seealso:: See :meth:`stop_server` for shutting down the adapter.
         """
-        pass
+        raise NotImplementedError(
+            'Adapters must implement start_server to construct and setup any servers or mechanism '
+            'required for network communication.')
 
     def stop_server(self):
         """
-        This method should be re-implemented to stop and tear down anything that has been setup
+        This method must be re-implemented to stop and tear down anything that has been setup
         in :meth:`start_server`. This method should close all connections to clients that have
         been established since the adapter has been started.
 
@@ -88,7 +90,9 @@ class Adapter(object):
             This method may be called multiple times over the lifetime of the Adapter, so it is
             important to make sure that this does not cause problems.
         """
-        pass
+        raise NotImplementedError(
+            'Adapters must implement stop_server to tear down anything that has been setup in '
+            'start_server.')
 
     @property
     def is_running(self):
--- a/lewis/adapters/__init__.py
+++ b/lewis/adapters/__init__.py
@@ -67,8 +67,38 @@ class Adapter(object):
         protocol in question. These startup operations are not supposed to be carried out on
         construction of the adapter in order to preserve control over when services are
         started during a run of a simulation.
+
+        .. note::
+
+            This method may be called multiple times over the lifetime of the Adapter, so it is
+            important to make sure that this does not cause problems.
+
+        .. seealso:: See :meth:`stop_server` for shutting down the adapter.
         """
         pass
+
+    def stop_server(self):
+        """
+        This method should be re-implemented to stop and tear down anything that has been setup
+        in :meth:`start_server`. This method should close all connections to clients that have
+        been established since the adapter has been started.
+
+        .. note::
+
+            This method may be called multiple times over the lifetime of the Adapter, so it is
+            important to make sure that this does not cause problems.
+        """
+        pass
+
+    @property
+    def is_running(self):
+        """
+        This property indicates whether the Adapter's server is running and listening. The result
+        of calls to :meth:`start_server` and :meth:`stop_server` should be reflected as expected.
+        """
+        raise NotImplementedError(
+            'Adapters must implement the is_running property to indicate whether '
+            'a server is currently running and listening for requests.')
 
     def handle(self, cycle_delay=0.1):
         """
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -79,6 +79,7 @@ class PropertyExposingDriver(Driver):
         self._target = target
         self._pv_dict = pv_dict
         self._timers = {k: 0.0 for k in self._pv_dict.keys()}
+        self._last_update_call = None
 
     def write(self, pv, value):
         pv_object = self._pv_dict.get(pv)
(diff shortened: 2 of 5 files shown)


lewis/adapters/__init__.py
```diff
@@ -67,9 +67,39 @@ def start_server(self):
...
+
+    @property
+    def is_running(self):
+        """
+        This property indicates whether the Adapter's server is running and listening. The result
+        of calls to :meth:`start_server` and :meth:`stop_server` should be reflected as expected.
+        """
+        raise NotImplementedError(
+            'Adapters must implement the is_running property to indicate whether '
+            'a server is currently running and listening for requests.')
```
MikeHart85, line 105: This appears to be the odd one out, raising a `NotImplementedError`. All the other methods here pass or do some default behaviour quietly. Is this intended?

lewis/adapters/epics.py
```diff
@@ -205,12 +206,17 @@ def start_server(self):
...
+            self._server.createPV(prefix=self._options.prefix,
+                                  pvdb={k: v.config for k, v in self.pvs.items()})
+            self._driver = PropertyExposingDriver(target=self, pv_dict=self.pvs)
+
+            self._last_update = datetime.now()
 
-        self._last_update = datetime.now()
+    def stop_server(self):
+        self._server = None
+        self._driver = None
```
MikeHart85, line 221: I'm not sure if it matters, but shouldn't this be the other way around? Destroy the driver first, then the server?

MikeHart85: On second thought... Will this not potentially be a problem in general? Completely destroying and reinitializing the server and driver (and all PVs along with them)?

Especially considering #169.

Won't this reset all runtime changes to PVs? If so, I think that would be very unexpected behaviour.

lewis/adapters/epics.py
```diff
@@ -236,6 +242,7 @@ def handle(self, cycle_delay=0.1):
 
         :param cycle_delay: Approximate time to be spent processing requests in pcaspy server.
         """
-        self._server.process(cycle_delay)
-        self._driver.process_pv_updates(seconds_since(self._last_update))
-        self._last_update = datetime.now()
+        if self._server is not None:
+            self._server.process(cycle_delay)
+            self._driver.process_pv_updates(seconds_since(self._last_update))
+            self._last_update = datetime.now()
```
MikeHart85: No `is_running` implementation for EPICS?

lewis/adapters/stream.py
```diff
@@ -78,13 +78,29 @@ def __init__(self, host, port, target):
...
+        self._accepted_connections = []
+
     def handle_accept(self):
         pair = self.accept()
         if pair is not None:
             sock, addr = pair
             print("Client connect from %s" % repr(addr))
             StreamHandler(sock, self.target)
 
+            self._accepted_connections.append(sock)
```
MikeHart85: `sock`s are appended to `accepted_connections` but only removed on server shutdown. What about connections closed by the client?

lewis/adapters/__init__.py
MichaelWedel: Yes, this is intended, because `Simulation` relies on this working correctly so I thought returning `False` or `True` by default could be a bit dangerous. With the exception it's ultimately clear that this needs to be done. I'm almost thinking we should maybe make `start_server` and `stop_server` behave the same way.

lewis/adapters/epics.py
MichaelWedel: I haven't tried in connection with #169, but in principle this uses the same mechanism as PV value updates (and on the same poll interval). But now that you mention it we should probably have one forced poll after instantiating the driver so that everything is updated. Since all that is happening in one cycle I don't think there will be a problem.

lewis/adapters/epics.py
MichaelWedel: Good point :)

lewis/adapters/stream.py
MichaelWedel: I tried manually by disconnecting and I couldn't see any adverse behavior (exceptions etc), so I assume calling `.close` multiple times doesn't matter for the socket. What I'm not sure about are the implications of keeping the object around.

We could also try storing the `StreamHandler` and calling `close` on that instead. That would not change the fact that we'd keep the handler around for longer than necessary.

lewis/adapters/epics.py
MichaelWedel: With the one forced update after starting it seems to work as expected. I also changed the order of "destruction" - although I'm not sure whether that changes anything either.

lewis/adapters/epics.py
```diff
@@ -205,12 +207,22 @@ def start_server(self):
...
-                              pvdb={k: v.config for k, v in self.pvs.items()})
-        self._driver = PropertyExposingDriver(target=self, pv_dict=self.pvs)
+        if self._server is None:
+            self._server = SimpleServer()
+            self._server.createPV(prefix=self._options.prefix,
+                                  pvdb={k: v.config for k, v in self.pvs.items()})
+            self._driver = PropertyExposingDriver(target=self, pv_dict=self.pvs)
+            self._driver.process_pv_updates(0.0, force=True)
+
+            self._last_update = datetime.now()
```
MikeHart85: While we're mucking around in here... `_last_update` could use some polish. 

It's defined outside of `__init__`. Also, it's only really user inside `PropertyExposingDriver.process_pv_updates()`. But defined here as a member of `EpicsAdapter`. It's all a bit awkward.

MichaelWedel: True, that looked really strange. I moved it to the driver, that looks better now.

PR #172, opened 2017-01-04
Add release notes to generated documentation

### LINKED ISSUE #168: Add release notes to generated documentation
As discussed on Slack, we should have release notes. Probably the easiest way is to simply make those part of the source code like all other parts of the documentation.

Each pull request should then add to the release notes if necessary.

MichaelWedel:
This fixes #168.

This adds release notes to the sphinx generated documentation so that it becomes part of the source and the online docs.

The current release has the `:orphan:` marker at the top to suppress a sphinx warning, because it is not included in the TOC yet, because it's work in progress and not yet intended to be seen by users.

After this has been merged, the current release document should be updated along with the code when changes are made.

### DIFF
--- a/docs/release_notes/release_1_0_0.rst
+++ b/docs/release_notes/release_1_0_0.rst
@@ -1,0 +1,23 @@
+Release 1.0
+===========
+
+The initial release of Lewis (at that point still plankton). These release notes have been
+compiled after the release as we were not keeping any release notes at that time.
+
+Features
+--------
+
+ - Cycle-based, deterministic device simulations based on finite state machines
+ - Control over the simulation's time granularity and speed (slow motion, fast forward)
+ - Simulation and device control via command line and via optional network service
+ - Two ready to use simulated devices using different protocols:
+
+    - ESS chopper abstraction (CHIC) using EPICS Channel Access
+    - Linkam T95 temperatur controller using TCP Stream protocol
+
+ - Docker image available on `docker hub`_ containing all dependencies
+ - Documentation in Markdown format for viewing in Github, both for users and developers
+ - Examples for implementing Stream protocol based devices
+
+.. _pcaspy: https://github.com/paulscherrerinstitute/pcaspy
+.. _docker hub: https://hub.docker.com/r/dmscid/lewis/
--- a/docs/release_notes/release_1_0_0.rst
+++ b/docs/release_notes/release_1_0_0.rst
@@ -13,7 +13,7 @@ Features
  - Two ready to use simulated devices using different protocols:
 
     - ESS chopper abstraction (CHIC) using EPICS Channel Access
-    - Linkam T95 temperatur controller using TCP Stream protocol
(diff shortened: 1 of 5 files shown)


docs/release_notes/release_1_1_0.rst
```diff
@@ -0,0 +1,20 @@
...
+------------
+
+ - Exposing devices via TCP Stream has been made easier. It is now possible to define commands
+   that expose lambda-functions, named functions and data attributes (with separate read/write
+   patterns).
+
+Bug fixes and other improvements
+--------------------------------
+
+ - Virtually disconnecting devices via the control server now actually closes all network
```
OwenArnold, line 18: Would be useful to either have some imbedded examples or links to specific documentation for these new features. You already did that for the `pip install` thing in the `1_0_1` release notes.

docs/release_notes/release_1_0_0.rst
```diff
@@ -0,0 +1,23 @@
...
+Features
+--------
+
+ - Cycle-based, deterministic device simulations based on finite state machines
+ - Control over the simulation's time granularity and speed (slow motion, fast forward)
+ - Simulation and device control via command line and via optional network service
+ - Two ready to use simulated devices using different protocols:
+
+    - ESS chopper abstraction (CHIC) using EPICS Channel Access
+    - Linkam T95 temperatur controller using TCP Stream protocol
```
OwenArnold: Minor spelling issue of temperature

OwenArnold:
Some minor updates to make. Documentation built and displayed as expected.

docs/release_notes/release_1_0_1.rst
```diff
@@ -0,0 +1,33 @@
...
+ - pcaspy is now an optional requirement that has to be enabled explicitly in the requirements.txt
+   file or installation via ``pip install lewis[epics]``.
+ - Error messages displayed on the command line have been improved.
+ - A flake8 job has been added to the continuous integration pipeline to enforce Python
+   style guidelines in the codebase.
+
+
+.. _David Michel: https://github.com/dmichel76
+.. _PyPI: https://pypi.python.org/pypi/lewis
+.. _RTD: http://lewis.readthedocs.io/en/latest/
```
MikeHart85: If possible, I would link this directly to the correct version rather than `latest`.

MikeHart85: Actually, in the context this is used in, latest probably makes more sense. But shouldn't we also include links to the correct version of documentation in all release notes?

MichaelWedel: Thought about this as well and had "stable" first instead of "latest" but that doesn't make sense either. We'd have to add a separate build for each specific release tag in RTD, but that should not be a problem.

MichaelWedel:
I added a few small things (no extensive examples, but a link to the relevant API documentation). There's also a `v1.0.1` build on RTD now and the link in the release notes is pointing to that build now.

PR #175, opened 2017-01-12
Support for Modbus Protocol

### LINKED ISSUE #160: Create a MODBUS interface
At ISIS we are trying to write an emulator for a Eurotherm nanodac http://www.eurotherm.co.uk/products/recorders/graphic/nanodac. We plan to talk to the device using MODBUS TCP and already have an IOC (written elsewhere). The use case for writing an emulator is that we have yet to purchase the equipment and we would like to test the IOC.

We are happy to do the actual device emulator ourselves but it would require MODBUS support within plankton.

MikeHart85:
Fixes #160.

This PR adds a Modbus Adapter and an example device that uses it.

Several Modbus libraries were considered, but all of them had issues that disqualified them. `pymodbus` was the most feature complete Modbus implementation and was very attractive. Unfortunately, it doesn't support Python 3 as of this writing. Consequently, we decided to implement a small subset of Modbus ourselves and expand it as required.

References used:
* http://www.modbus.org/docs/Modbus_Application_Protocol_V1_1b3.pdf
* http://www.modbus.org/docs/Modbus_Messaging_Implementation_Guide_V1_0b.pdf
* https://github.com/sourceperl/pyModbusTCP
* https://github.com/bashwork/pymodbus

Currently, this supports the following Modbus Function Codes:
* READ_COILS = 0x01
* READ_DISCRETE_INPUTS = 0x02
* READ_HOLDING_REGISTERS = 0x03
* READ_INPUT_REGISTERS = 0x04
* WRITE_SINGLE_COIL = 0x05
* WRITE_SINGLE_REGISTER = 0x06
* WRITE_MULTIPLE_COILS = 0x0F
* WRITE_MULTIPLE_REGISTERS = 0x10

A preliminary `ModbusDataBank` is provided, which only supports providing a contiguous memory range for Modbus to read from and write to. Binding memory addresses to `Device` or `Interface` attributes or functions is not supported yet. The current implementation is meant to serve as a foundation for future additions, as they become required. The idea is that you will always be able to create a simple DataBank such as this, by doing, e.g.:
```python
di = ModbusDataBank.create_basic(False, 0x1000, 0x1FFF)
```
But, in addition, more complex DataBanks, including ones that bind to attributes and functions, could be created with different `create_` methods. Using the `__init__` constructor directly should be avoided.

Overlaid memory banks may be simulated by simply assigning the same DataBank instance to multiple DataStore segments:
```python
di = ModbusDataBank.create_basic(False, 0x1000, 0x1FFF)
co = di
```
This means Coils and Discrete Inputs will interact with the same memory, mapped 1 to 1. Here, ideas for future work, in case more complex mapping scenarios are required, would be to allow one DataBank to be mapped to another using wrapper classes or a `create_mapped` method, for example.

Attempts to read from or write to memory locations outside of the specified limits causes an IndexError exception to be raised. Similarly, Modbus requests for addresses that fall outside of the range cause a DATA_ADDRESS exception packet to be sent back. 

Values written to the DataBank by Modbus are always `bool` for coils/inputs and (16-bit unsigned range) `int` for registers. From the code, any values may be written the the DataBank, but they will be cast to `bool` or truncated to 16 bit `int` when read by Modbus.

## Running

Example device can be run like this:
```
$ ./lewis.py -k lewis.examples modbus_device -- -b localhost -p 5020
```
This will listen on `localhost:5020`. The adapter default to port 502, as required by Modbus specification.

For testing, I used `pymodbus` client examples, which can be found here:
https://pymodbus.readthedocs.io/en/latest/examples/synchronous-client.html
https://pymodbus.readthedocs.io/en/latest/examples/asynchronous-client.html
Make sure to change the port number to 5020 if you want to run them.


MikeHart85:
We need to figure out good way to bind the adapters to the interface (the way the EPICS Adapter provides `PV` and Stream Adapter provides `Cmd` and `Var`) that provides a clean usage and is flexible enough to support differences between devices. 

This will likely be an ongoing process as we add more devices (although it would be good to think ahead and discuss here), as with the other adapters, so this PR will focus on providing @DominicOram with what he needs for the `nanodac`.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+        hexdata = str([c.encode('hex') for c in data])
+        print(">>> " + hexdata)
+        self.parser.parse(data)
+
+    def handle_close(self):
+        print("Client from %s disconnected." % repr(self.addr))
+        self.close()
+
+
+class ModbusServer(asyncore.dispatcher):
```
MichaelWedel, line 520: This class looks a lot like `StreamServer`, with the exception of the handler. Maybe we could extract this into a base class and just make the handler a class variable that can be overridden. There's also some more code in `StreamHandler` for dealing with connect/disconnect.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+        self._server = None
+
+        if self.databank is None:
+            raise RuntimeError("Must specify a databank to use ModbusAdapter.")
+
+    def _parse_arguments(self, arguments):
+        return {}
+
+    def start_server(self):
+        self._server = ModbusServer('localhost', 5020, self)
```
MichaelWedel: Are host and port eventually envisioned to be adapter arguments?

lewis/examples/modbus_device/__init__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# *********************************************************************
+# lewis - a library for creating hardware device simulators
+# Copyright (C) 2016 European Spallation Source ERIC
```
MichaelWedel, line 4: This should be 2017 now - or 2016 - 2017? I vaguely remember that we talked about this at some point.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+        if pair is not None:
+            sock, addr = pair
+            print("Client connect from %s" % repr(addr))
+            ModbusHandler(sock, self.target)
+
+    def handle_close(self):
+        pass
+
+
+class ModbusAdapter(Adapter):
```
MichaelWedel, line 547: Similar to `ModbusServer`, this looks very much like `StreamAdapter`. I suspect the TCP/IP-nature of the protocols leads to this.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+        addr, value = struct.unpack('>HH', bytes(self._request.data))
+
+        if not self.databank.validate_words(addr):
+            self._exception = MBEX.DATA_ADDRESS
+            return
+
+        self.databank.set_words(addr, [value])
+        self._response = self._request.create_response()
+
+        print("Write WORD request for value {} at address{}".format(value, addr))
```
MichaelWedel: We'll need to make sure to replace these prints with logging statements once this is introduced.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+            self._exception = MBEX.DATA_VALUE
+            return
+
+        if not self.databank.validate_bits(addr, count):
+            self._exception = MBEX.DATA_ADDRESS
+            return
+
+        # Get response data
+        bits = self.databank.get_bits(addr, count)
+        byte_count = int(ceil(len(bits) / 8.0))
```
MichaelWedel: Just a little detail, have you considered `from __future__ import division`? Then it would be possible to write `len(bits) / 8`. It won't change the functionality, but somehow "8.0 bits" does sound a bit odd.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+class ModbusBasicDatabank(object):
+    def __init__(self):
+        self.bits = [False] * 0x10000
+        self.words = [0] * 0x10000
+
+    def get_bits(self, address, count=1):
+        bits = self.bits[address:address+count]
+        return bits if len(bits) == count else None
```
MichaelWedel: As a general question, is there a particular reason why these return `None` instead of raising an exception?

lewis/examples/modbus_device/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# *********************************************************************
+
+from lewis.devices import Device
+from lewis.adapters.modbus import ModbusAdapter, ModbusBasicDatabank
+
+
+class ModbusDevice(Device):
```
MichaelWedel, line 24: I think you beat the `SimpleDevice` in terms of simplicity :)

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def parse(self, data=None):
+        """
+        Parse as much of current buffer as possible.
+
+        :param data: Optionally append given data to buffer prior to parsing
+        """
+        self.buffer.extend(bytearray(data or ''))
+
+        # Process until two or three FSM cycles stay in the same state
```
MichaelWedel: Why is this required? Could the protocol have something like a `parse_complete` state?

lewis/core/statemachine.py
```diff
@@ -294,6 +295,8 @@ def bind_handlers_by_name(self, instance, override=False, prefix=None):
         """
         if prefix is None:
             prefix = {}
+        if not isinstance(prefix, dict) and hasattr(prefix, '__iter__'):
```
MichaelWedel, line 307: Hm...it really needs to be a list or some other ordered collection, right? If someone passed a set, unexpected things would most likely happen. I don't have a suggestion for a better check at the moment, but I think this might be too forgiving. Maybe it's even okay to demand a dict here (as it was before) to eliminate any ambiguity, even if it's slightly more verbose on the caller's end.

lewis/core/statemachine.py
```diff
@@ -294,6 +295,8 @@ def bind_handlers_by_name(self, instance, override=False, prefix=None):
...
+            prefix = dict(zip(['on_entry', 'in_state', 'on_exit'], prefix))
 
         # Merge prefix defaults with any provided prefixes
         prefix = dict(list(self._prefix.items()) + list(prefix.items()))
```
MichaelWedel, line 311: Could this be a candidate for `dict_strict_update` while we're doing changes in this general area?

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,450 @@
...
+    SLAVE_DEVICE_BUSY = 0x06
+    MEMORY_PARITY_ERROR = 0x08
+    GATEWAY_PATH_UNAVAILABLE = 0x0A
+    GATEWAY_TARGET_DEVICE_FAILED_TO_RESPOND = 0x0B
+
+
+class ModbusBasicDatabank(object):
+    def __init__(self):
+        self.bits = [False] * 0x10000
+        self.words = [0] * 0x10000
```
MichaelWedel: As far as I understood from briefly looking at the modbus specification PDF you linked words are 16 bit integers. Could `array` (https://docs.python.org/2/library/array.html) be a way to make this restriction more visible?

MichaelWedel (commented):
I like it, very structured approach to parsing the protocol, although I haven't had time to look at the details of the parser. What I definitely like is that this is a very different use of `StateMachine`.

Regarding the "bindings", do you think it would be reasonable to have something like `Word`, `Coil`, `Register` (as appropriate) that bind directly to the device's attributes?

lewis/adapters/modbus.py
MikeHart85: I'm not sure if this applies to Modbus over TCP, but in general there is no guarantee in the protocol that requests arrive as one unit. This is to ensure ModbusProtocol supports getting incomplete chunks of requests.

When I started, I was considering using StateMachine to parse the incoming stream byte by byte (or word by word). 

But in the end, it seemed far simpler to encapsulate reading in ModbusTCPFrame and just use ModbusProtocol and its StateMachine to step through the higher level protocol exchange. Not entirely sure using StateMachine is even of advantage now... the only real branching point is after checking the function code, and that could be done with a function table.

lewis/adapters/modbus.py
MikeHart85: I think at one point I was going to use this for validation. Later, I added the `validate_*` methods. Need to clean this up a bit.

Not sure how the StateMachine would respond to having an exception thrown here. I don't think it would catch it, so it would probably bring down the Adapter/Lewis as a whole. We might want to consider if it would make more sense for StateMachine to catch exceptions that happen in state handlers.

lewis/adapters/modbus.py
MikeHart85: Sounds like a good idea, I'll have a look. Thanks!

This basic databank is pretty much copied straight from pyModbusTCP and just there as a placeholder, it need a lot more work. Especially since it needs to interact with the Interface/Device.

lewis/adapters/modbus.py
MikeHart85: Ah, cool, didn't know about that. Yes, that's perfect.

lewis/adapters/modbus.py
MikeHart85: Yes, was just focused on the Modbus protocol to start with.

lewis/examples/modbus_device/__init__.py
MikeHart85: Doing some research on this:
* Copyright notices are generally optional and purely educational (no bearing on legal protection)
* The year should be the first year of publication of a work
* Optionally, you may update the year (or list years) for any year with significant changes made
* GNU recommends: `2008, 2009, 2010` or `2008-2010` format

For purposes of software, a "work" is the entire software package as a whole, not individual files.

Long story short: 
We can either leave it `2016` everywhere and just not worry about it, or, if we want to update, we should search replace the entire source folder to `2016-2017` or similar. Best to do it in one go as a PR if we go that route, I think.

MikeHart85: https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

lewis/examples/modbus_device/__init__.py
MikeHart85: Even beats `VerySimpleDevice` :smile:

lewis/core/statemachine.py
MikeHart85: A `set` should work fine from what I can tell. It's just an unordered, unique-only list really. The only unexpected behaviour would arise from the unordered nature of it... but that's just you getting exactly what you asked for by using a `set`. :smile: 

Further, `isinstance(OrderedDict(), dict)` returns `True`, so there shouldn't be any issues with those kinds of collections either.

lewis/core/statemachine.py
MikeHart85: Hm. In principle yes, but I'm a bit concerned that this isn't currently `self._prefix.update(prefix)`... is it really functionally equivalent?

MikeHart85: Actually, it looks like `self._prefix` is only ever used in this function... was this always the case? I could have sworn it had more uses.

MichaelWedel: It's probably still better to not modify it, because that will change the behavior when calling the function multiple times. I was just thinking that it might be good to use `dict_strict_update` to catch key-typos in a dict that is passed in.

lewis/adapters/modbus.py
MikeHart85: Ah, I think I misunderstood the question a bit (thought it was about the buffer)... but still sort of answered it, partially, in a roundabout way.

Now that `ModbusTCPFrame` always returns a complete request, the condition this is really waiting for is to "spend more than one cycle in `await_request`". Because this means the buffer no longer contains any complete requests (and we can't do anything anymore except keep waiting).

This is necessary because sometimes more than one request comes in as one big chunk of requests. If you try using the asynchronous client example, instead of the synchronous one, you'll see that ALL the requests come in as one big packet in that one. I haven't seen partial requests yet, but in principle this ModbusAdapter would support that.

I wouldn't really want a `parse_complete` state since... well, what do you do in that? If you're done with one request, you should wait for the next one. Or go ahead and process the next one, if you already have enough data in your buffer for it.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,405 @@
...
+        """Return handler with signature handler(request) for function code fcode"""
+        return self._handlermap.get(
+            fcode,
+            lambda req: req.create_exception(MBEX.ILLEGAL_FUNCTION)
+        )
+
+    def _handle_read_coils(self, request):
+        """
+        Handle request as per Modbus Application Protocol v1.1b3:
+        Section 6.1 - (0x01) Read Coils
```
MikeHart85, line 324: Not sure if this is the best way to reference a section in the PDF (particularly considering Sphinx/ReadTheDocs). Anyone have ideas on how to best format/word this?

Should also consider if we can / want to keep a copy of that document in our repo and/or documentation.

DominicOram:
Sorry I haven't had much of a chance to look at this yet. I have managed to get the nanodac IOC communicating with the emulator and from the emulator output the IOC appears to be doing roughly what I would expect. However, when I write and subsequently read from a register location I do not see my written value. I haven't ruled out a problem with the IOC itself yet and will continue looking later this week.

MikeHart85:
That was my fault! Lazy copy-paste... from already copy-pasted code, no less! 

Reading registers should work now.

DominicOram:
Great, that all works as expected now. To be honest we're using the nandoc for so little (only basic reading and writing) that the simple device may be sufficient. However, if we start to find any oddities when we get the actual kit I will write a nanodac device. Thank you for all your work, I'm happy for this to be merged.

MikeHart85:
Glad that works for you, I had a feeling from our discussions that the nanodac wouldn't require much more. I'll still flesh out the adapter a bit more (so things like hostname and port can be changed) and maybe improve the way the databank works a bit before merging. We can add more feature as they become needed later.

MikeHart85:
Should be ready for merge pending final review / testing.

I decided against implementing attribute and function bindings at this stage. It just feels like there are too many unknowns/questions about what would work best to model device behaviour in a practical way. It would probably be better to wait for a more complex Modbus device, so that we have something concrete to design for and test against.

I did rewrite `ModbusDataBank` a bit to be more generic and introduced `ModbusDataStore`. So a DataBank is now a fairly abstract representation of a chunk of memory that can be used to store and read values (whether they are bits or words doesn't matter). The DataStore consists of four references to DataBanks, which are used for Discrete Inputs, COils, Input Registers and Holding Registers, respectively.

As @MichaelWedel noted, there is significant overlap between the `modbus` and `stream` adapters when it comes to the TCP handling. Factoring this out is left as an exercise for another PR.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,569 @@
...
+
+            di = ModbusDataBank.create_basic(False, 0x1000, 0x1FFF)
+
+        :param config: dict containing configuration
+        """
+        self._data = config['data']
+        self._start_addr = config['start_addr']
+
+    @classmethod
+    def create_basic(cls, default_value=0, start_addr=0x0000, last_addr=0xFFFF):
```
MichaelWedel: Why is this implemented as a factory method? We haven't used this pattern anywhere else so far, could this be implemented with a `ModbusBasicDataBank` that calls the base class `__init__` method with the proper parameters? That would achieve the same thing (shield the interface of `ModbusDataBank.__init__` from the user) but a bit more in line with how we create things in other places in the framework.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,569 @@
...
+    DATA_VALUE = 0x03
+    SLAVE_DEVICE_FAILURE = 0x04
+    ACKNOWLEDGE = 0x05
+    SLAVE_DEVICE_BUSY = 0x06
+    MEMORY_PARITY_ERROR = 0x08
+    GATEWAY_PATH_UNAVAILABLE = 0x0A
+    GATEWAY_TARGET_DEVICE_FAILED_TO_RESPOND = 0x0B
+
+
+class ModbusTCPFrame(object):
```
MichaelWedel, line 133: You already mentioned that we need to refactor the Modbus and Stream infrastructure in order to minimize code duplication. Maybe TCP Stream could profit from something similar to this, although it might be harder to do this for "free format protocols". Something to think about for the next release.

lewis/adapters/modbus.py
MichaelWedel: I think for now it's okay, there could be a link to the PDF or the/a page where the document can be obtained.

lewis/adapters/modbus.py
```diff
@@ -0,0 +1,569 @@
...
+        pair = self.accept()
+        if pair is not None:
+            sock, _ = pair
+            handler = ModbusHandler(sock, self.adapter, self)
+            self._accepted_connections.append(handler)
+
+    def remove_handler(self, handler):
+        self._accepted_connections.remove(handler)
+
+    def handle_close(self):
```
MichaelWedel, line 540: In `StreamServer` this is done by overriding `close`, we have to harmonize this in the next release. As long as it's functionally equivalent, it's probably okay.

lewis/examples/modbus_device/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+
+    Here, two basic ModbusDataBanks are created and initialized to a default value across the full
+    range of valid addresses. One DataBank is shared by di and co, and the other by ir and hr to
+    demonstrate overlaid memory segments. If you want each segment to have its own memory, just
+    create separate instances for all four.
+    """
+    di = ModbusDataBank.create_basic(False)
+    co = di
+    ir = ModbusDataBank.create_basic(0)
+    hr = ir
```
MichaelWedel, line 42: I have the feeling that we should be very, very careful with labeling this as a general example for a Modbus device given that there is basically no support for exposing actual device behavior yet. Correct me if I'm wrong, but this is basically the only possible device at the moment, right? The only reason it's required there is that we once said "A Simulation requires a device".

Could we maybe move this to `devices` and call it something like `modbus_echo` or something like that? Combined with the logging functionality it would be very useful for diagnosing IOC-level problems concerning correct addresses and transmitted values and so on - and it would never change.

But once there are binding tools like for the other protocols, this will change significantly, so from my point of view it would be better to not put this into the "examples" module.

MichaelWedel (commented):
`ModbusProtocol` has become significantly more simple, even though I liked the new use of `StateMachine` - this is clearly the better way to go, very nice.

I would like to carry out some functional tests myself, although @DominicOram has apparently already done so extensively.

lewis/adapters/modbus.py
MikeHart85: That would work as well... But I imagine we will eventually wind up with many tiny classes that are almost identical, for the various scenarios. 

Inheritance makes most sense when you have a common interface but differing implementations. In this case, I think both interface and implementation can/will always be the same. But the initialization will differ depending on what kind of DataBank you want. Consider:

```python
@classmethod
def create_remapped(cls, databank, new_start_addr):
    return cls({'start_addr': new_start_addr, 'data': databank._data})
```
```python
di = ModbusDataBank.create_basic(False, 0x1000, 0x1FFF)
co = ModbusDataBank.create_remapped(di, 0x2000)
```
Now `co` addresses 0x2000 - 0x2FFF map to the same memory as `di` address 0x1000 - 0x1FFF. (Looking at this now, I should probably have used args/kwargs in` __init__` rather than a single dict parameter; would make the syntax in the classmethods clearer.)

You could also have various types of conversion between values and address spaces of bit-based DataBanks and word-based DataBanks. DataBanks that bind to attributes / functions would have `create_bound` or similar. The internal implementation of ModbusDataBank (including `get` and `set`) would of course have to be extended significantly to support all this. 

One advantage of doing it this way is that it emphasizes that a DataBank is always fundamentally the same type of object; all that changes is how it is set up.

But I'm not very attached to this pattern, we could certainly achieve the same with inheritance. Just felt like the natural way to go... I think I saw something similar used in pymodbus somewhere and thought "ah, we could do even more with this".

You're right it's very different from what we've done in other places. I'm happy to change it if you think it's too far from our usual MO.

lewis/adapters/modbus.py
MikeHart85: Yeah, I'm not sure which way is more correct... or whether there's any difference at all. Agreed, something to look at when we refactor here. Best to leave as is to serve as a reminder that this is something to look into.

MichaelWedel: I have a feeling that `handle_close` might be the right method to override.

lewis/examples/modbus_device/__init__.py
MikeHart85: It's not quite the only possible device. This would behave differently (DataBank memory space not shared):
```python
class ModbusInterface(ModbusAdapter):
    di = ModbusDataBank.create_basic(False)
    co = ModbusDataBank.create_basic(False)
    ir = ModbusDataBank.create_basic(0)
    hr = ModbusDataBank.create_basic(0)
```
And so would this (limited address ranges, requests to outside range return Modbus exception packet):
```python
class ModbusInterface(ModbusAdapter):
    di = ModbusDataBank.create_basic(False, 0x1000, 0x1FFF)
    co = di
    ir = ModbusDataBank.create_basic(0, 0x1000, 0x1FFF)
    hr = hr
```
So while it's not a feature complete example (as far as features we plan to implement but haven't yet), it is an example rather than a generic thing onto itself.

I can see a `modbus_echo` being useful... but which version of the above should it look like? I think in practice you might want to just take this example and tweak it for what you need at the moment.

lewis/adapters/modbus.py
MichaelWedel: > The internal implementation of ModbusDataBank (including get and set) would of course have to be extended significantly to support all this.

This sounds like an argument for sub-classing to me. I like your idea of using `kwargs` for the base `__init__`, it might simplify change in the future as well.

So how about something like this?

```python
class ModbusDataBank(object):
    def __init__(self, *args, **kwargs):
        self.data = kwargs['data']
        self._start_addr = kwargs['start_addr']

    # and so on

class ModbusBasicDataBank(ModbusDataBank):
    def __init__(default_value=0, start_addr=0x0000, last_addr=0xFFFF):
        super(ModbusBasicDataBank, self).__init__(start_addr=start_addr, data=[default_value] * (last_addr - start_addr + 1))
```

That way `ModbusDataBank` does not need to change when new implementations are added. The downside is you'd need to know what sub-classes exist in order to import them to implement your interface, but this way it's also easier to draw a line and say "don't use this directly, use one of the sub-classes instead".

lewis/examples/modbus_device/__init__.py
MichaelWedel: Thanks for showing some more examples, it's more clear now. I guess what I meant in my comment above is that is bugs me slightly is that it is still the same `Device`, just the `Interface` changes.

Regarding your question which version it should look like, I think we're getting closer. Could `di`, `co`, `ir`, `hr` exist in the `Device` as well? Because then I could imagine the first proper use of setups beyond having the possibility to showcase them in one of our examples :)

Consider something like `StateMachineDevice`:

```python

class ModbusEchoDevice(Device):
    def __init__(self, di, co, ir, hr):
        self.di = di
        self.co = co
        self.ir = ir
        self.hr = hr


setups=dict(
    default=dict(
        device_type=ModbusEchoDevice,
        parameters=dict(
            di=ModbusDataBank.create_basic(False, 0x1000, 0x1FFF),
            ...
        )
    )
    non_overlapping=....
)

Then my answer to your question which one of the above would be "all of them" (not quite, but some useful ones).

MikeHart85: Yes, it's weird that the `Device` is just a `pass`. But that's just a consequence of not having bindings yet. We can just update the example once we implement those.

> Could `di`, `co`, `ir`, `hr` exist in the `Device` as well?

I had a similar thought at some point, but dismissed it because this would break our model of system separation, no? `Device` is supposed to be just a Python class that knows nothing of protocol or adapters, and `Interface` acts as the binding between `Adapter` and `Device`?

MichaelWedel: Yes, it would break our separation, that is true, so maybe not a good way to go after all. I think however that it urgently shows the need for proper bindings, which brings me back to the original comment...how do we actually advertise this until we have such bindings?

In the generated documentation (the `rst`-files still need to be added under `docs/api/adapters` so that this gets generates) this should be marked as experimental with a small section describing how this works for now. One paragraph with "this is how you could modify the `Interface` in the example to achieve X instead" would probably enough.

Speaking of documentation, the release notes should also be updated. I also keep forgetting it as the procedure is new :)

MikeHart85: Well, without an example device and someone to ask questions about it, any sort of bindings I could implement now would be based entirely on speculation about how I _think_ a modbus device _might_ work (based mostly on how EPICS and TCP Stream devices work). The `nanodac` (so far) only requires this sort of read/write memory functionality. So I'm not sure "urgently" is quite right... Even if we had bindings, we would have no device to use them, which is the reason we don't have bindings to begin with. It's a chicken or the egg type of problem.

Basically, it boils down to: We had no modbus adapter, so Dom complained and we implemented what he needed. Now we have a simple modbus adapter without bindings. We can either A) Wait for someone to complain they're missing and implement them then, or B) Implement speculative bindings and wait for someone to complain they're bad and implement them properly then. I'm not sure which would be better.

Hypothetically, if you were desperate enough, you could implement device behaviour by overriding `handle` in the interface and updating the DataBanks against the Device, and vice versa, every cycle. But that isn't really the "proper" way of doing things, and so I didn't want to encourage it just to make the example a bit more interesting. Seemed better to just show existing functionality and leave future functionality to the future.

Good point about the docs and release notes, thanks for the reminder!

lewis/adapters/modbus.py
MikeHart85: >This sounds like an argument for sub-classing to me.

Only if they need to do different things for different DataBanks. I don't think that would be the case. But `_data` would have to become a `dict` instead of a `list` and so `get` and `set` would have to change accordingly. Right now, this just isn't necessary.

But, yes, exactly the same can be achieved with inheritance... it's just a question of whether you have many subclasses or many `create_type` methods. Subclassing does have the (huge) advantage that you have the option of changing behaviour completely if you need to... the `create` methods just felt like a more tidy alternative.

lewis/adapters/modbus.py
MikeHart85: Thinking about this a bit, I don't think it will be possible, or at least very inconvenient. As you mention, TCP Stream is very free format... the only thing separating messages is the line terminator (which can vary). And what the message itself is structured like is entirely up to the device.

With Modbus, you have a header that always looks the same and always includes the length of the full frame. This is what makes cutting the incoming data stream into frames so clean and convenient. And they always work the same regardless of device.

PR #177, opened 2017-01-17
Fix display of unserializable properties in control script

### LINKED ISSUE #174: Non-serializable members cause control client failure
This is an issue reported by @AdrianPotter.

If a member attribute of the `Device` class is not serializable, attempting to access the list of Properties and Methods of the Device fails to list any members and throws an exception. Expected behaviour would be to skip (or issue a warning about) the offending member and still list the others.

To reproduce easily, just add a `bad_object = set()` to any existing device and try accessing it via control client, e.g.:

```python
class VerySimpleDevice(Device):
    param = 10
    bad_object = set()
```
```
$ ./lewis.py -r localhost:10000 -k lewis.examples simple_device
$ ./lewis-control.py device
  File "./lewis-control.py", line 24, in <module>
    control_simulation()
  File "lewis/scripts/control.py", line 107, in control_simulation
    show_api(remote, args.object)
  File "lewis/scripts/control.py", line 45, in show_api
    member = getattr(obj, member_name)
  File "lewis/core/control_client.py", line 215, in getter
    return obj._make_request(property_name + ':get')
  File "lewis/core/control_client.py", line 198, in _make_request
    raise exception(exception_message)
TypeError: [u'set([]) is not JSON serializable']
```
The error does not occur if the member is protected (`_bad_object`), so that can be used as a workaround. But, ideally, I think the control client / server should be smart enough to deal with this in a more user-friendly way. Perhaps just something like:

```
$ ./lewis-control.py device
Type: VerySimpleDevice
Properties:
	param
	bad_object >>> TypeError: [u'set([]) is not JSON serializable']
Methods:
	process
```

That way there's a good indication what works and what doesn't, and why.

MichaelWedel:
This fixes #174.

The interface is now correctly displayed also when certain members of the exposed object are not serializable. To make interaction easier, the current value of all device properties is printed, or a notice that it doesn't work and why.

### DIFF
--- a/lewis/scripts/control.py
+++ b/lewis/scripts/control.py
@@ -41,16 +41,20 @@ def show_api(remote, object_name):
     methods = []
 
     for member_name in dir(obj):
-        if not member_name[0] in ('_', ':'):
-            member = getattr(obj, member_name)
-
-            if callable(member):
-                methods.append(member_name)
+        if member_name[0] not in ('_', ':') and member_name not in dir(type(obj)):
+            methods.append(member_name)
 
     print('Type: {}'.format(type(obj).__name__))
-    print('Properties:')
+
+    print('Properties (current values):')
     for prop in sorted(properties):
-        print('\t{}'.format(prop))
+        try:
+            current_value = getattr(obj, prop)
+        except Exception as e:
+            current_value = 'Not accessible: {}'.format(e.message)
+
+        print('\t{}\t{}'.format(prop, current_value))
+
     print('Methods:')
     for method in sorted(methods):
         print('\t{}'.format(method))
--- a/lewis/scripts/control.py
+++ b/lewis/scripts/control.py
@@ -47,17 +47,18 @@ def show_api(remote, object_name):
     print('Type: {}'.format(type(obj).__name__))
 
     print('Properties (current values):')
+    maxlen = len(max(properties, key=len))
     for prop in sorted(properties):
         try:
             current_value = getattr(obj, prop)
         except Exception as e:
             current_value = 'Not accessible: {}'.format(e.message)
 
-        print('\t{}\t{}'.format(prop, current_value))
+        print('    {}    ({})'.format(prop.ljust(maxlen), current_value))
 
     print('Methods:')
     for method in sorted(methods):
-        print('\t{}'.format(method))
+        print('    {}'.format(method))
 
 
 def convert_type(value):


lewis/scripts/control.py
```diff
@@ -41,16 +41,20 @@ def show_api(remote, object_name):
     methods = []
 
     for member_name in dir(obj):
-        if not member_name[0] in ('_', ':'):
-            member = getattr(obj, member_name)
-
-            if callable(member):
-                methods.append(member_name)
+        if member_name[0] not in ('_', ':') and member_name not in dir(type(obj)):
+            methods.append(member_name)
```
MikeHart85, line 45: Won't this classify...
```python
class Foobar:
    hello = 'world'
```
...as a method since it's in `dir(type(obj))`? Or is this the client side "fake" `obj` that will look different?

MikeHart85: Wait... it's a `not in`... won't that miss members functions since they ARE in `dir(type(obj))`

MichaelWedel: The proxies on the client side are quite uniform. Plain attributes of server side objects are always properties, so they are in `dir(type(obj))`. Methods on the other hands are only put onto the object itself, so they are not in `dir(type(obj))`.

It might be slightly different on the server side, but the client does not know that currently.

MichaelWedel: Actually, to be a bit more precise, this is where the proxy objects are not always 100% accurate representations of the server side objects. I guess it's a trade-off between ease of implementation and difficulty to tell that it's a proxy object. At least at the moment I think this is acceptable, because you have to go and get the object explicitly anyway, so it's pretty clear that it's a server side object. The JSON serialization poses way more constraints anyway (but again, for the scope this is intended I find that acceptable).

lewis/scripts/control.py
```diff
@@ -41,16 +41,20 @@ def show_api(remote, object_name):
...
+
+    print('Properties (current values):')
     for prop in sorted(properties):
-        print('\t{}'.format(prop))
+        try:
+            current_value = getattr(obj, prop)
+        except Exception as e:
+            current_value = 'Not accessible: {}'.format(e.message)
+
+        print('\t{}\t{}'.format(prop, current_value))
```
MikeHart85: Consider for this segment:
```python
    print('Properties (current values):')
    maxlen = len(max(properties, key=len))
    for prop in sorted(properties):
        try:
            current_value = getattr(obj, prop)
        except Exception as e:
            current_value = 'Not accessible: {}'.format(e.message)

        print('    {}    ({})'.format(prop.ljust(maxlen), current_value))

    print('Methods:')
    for method in sorted(methods):
        print('    {}'.format(method))
```
Changes are:
- `ljust` to align columns
- parentheses around value to mirror "Properties (current values):" header
- four spaces instead of tabs for less width and more consistency (length of tab depends on where you tab from)

MichaelWedel: It really looks better and more consistent this way. Thanks!

MikeHart85:
Looks good, merging!

PR #178, opened 2017-01-19
Add logging capabilities

### LINKED ISSUE #79: Replace print statements with proper logging functionality
In some places, such as the EPICS adapter, there are still print statements. This should be replaced by proper logging functionality which can ideally at some point also be configured to log into files and other sinks, and also adjust the log level.

What are some nice packages we could use for this?

MichaelWedel:
This fixes #79.

This adds logging capabilities to the framework. There is now a new module `lewis.core.logging` which mostly does one thing, it defines a decorator called `has_log`. Applying that to a class (or a function actually) adds an attribute `log` which is a logger with a properly configured name. Because we use some classes in different places, it's possible to change the "context" of the logger, which essentially changes the name of the logger to make it clear which device something belongs to.

I tried doing this with inheritance first, but it did not look pretty, and there was the issue with mixing new- and old-style classes. Furthermore, random stuff would just end up inheriting from that logging mixin, which did not look nice either. It also prohibited the use in free functions (which is still sort of awkward, but I don't expect this to happen often).

Another advantage of this decorator based approach is that it's straight forward to add a note to the docstring of a decorated thing to include a note that this thing supports logging - the decorated class/function does not need to do anything for this.

All logger names are prefixed with `lewis.`. This may seem redundant, but since it is possible to use Lewis as a library, it must be possible for client code with own plans about logging to silence all of the library's logging output. This is possible by silencing the `lewis` logger, to which all loggers `lewis.x` forward their messages.

`lewis` has a new option `-o/--output-level` which gives control over the log level. Logging can also be disabled entirely by specifying `'none'`.

Currently this logs to `stderr` which I think is okay for the first iteration. I suppose we'll find issues over the next release cycle. By that time we'll also have more experience with the logging module.

The documentation has also been updated, so have the release notes.

### DIFF
--- a/lewis/core/logging.py
+++ b/lewis/core/logging.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # *********************************************************************
 # lewis - a library for creating hardware device simulators
-# Copyright (C) 2017 European Spallation Source ERIC
+# Copyright (C) 2016 European Spallation Source ERIC
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
--- a/lewis/core/logging.py
+++ b/lewis/core/logging.py
@@ -29,15 +29,18 @@ import logging
 
 class HasLog(object):
     log = None
+    logger_name = None
 
     def __init__(self, log=None):
         super(HasLog, self).__init__()
         self.logger_name = self.__class__.__name__
 
-        self.attach_log(log if log is not None else self.__class__.__name__)
+        self.attach_log(log)
 
     def attach_log(self, log):
         if log is not None:
             extension = log if isinstance(log, string_types) else log.__class__.__name__
 
             self.log = logging.getLogger('.'.join((extension, self.logger_name)))
+        else:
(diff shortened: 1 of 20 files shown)


lewis/core/logging.py
```diff
@@ -0,0 +1,141 @@
...
+
+        :param context: String or object, ``None`` to clear context.
+        """
+        obj.log.name = get_logger_name(context)
+
+    target.log = logging.getLogger(get_logger_name())
+    target._set_logging_context = _set_logging_context
+
+    if target.__doc__ is not None:
+        target.__doc__ += '''
```
MichaelWedel: Interesting, this is not allowed in Python 2, did not realize that. There are two ways around this, the easy one being to change this into a try...except block and configuration in RTD and build with Python 3. This will mean that documentation generated with Python 2 will look different.

There's also the possibility to change this into something slightly more complicated when `target` is a class, I'll play around with that a bit, but let me know if you think the configuration change route would be acceptable.

MikeHart85:
As discussed on Skype, looks good... merging!

PR #181, opened 2017-01-25
Put Modbus reference links into generated docs

MichaelWedel:
This (very small) change moves the links from the comment in modbus.py to the module's docstring so that they appear in the generated documentation.


MikeHart85:
Thanks, completely forgot to move those! :+1:

PR #182, opened 2017-01-25
Set multiple device parameters at once

### LINKED ISSUE #143: Setting multiple properties at the same time
A question that came up during a meeting with the IBEX team is whether multiple values could be set at once using `control.py` from the command line. 

My initial response was that this was a scenario where it might be more appropriate to write a script rather than using the command line utility.

Thinking about it, however, there may be scenarios where it is important to set multiple values atomically. That is, to ensure they are all set during the same cycle. 

Right now, due to the way the processing loop works, I think only one RPC command can be processed per cycle? If so, setting 10 different variables would get split across 10 cycles (1 second by default) even if you do it using a script. Even if this is not the case, it's impossible to ensure multiple properties are set during the same cycle.

We should consider:

* In the control client, making it possible to set multiple properties in one go using a dictionary or similar
* Exposing the above in the command line interface somehow (bonus, non-essential)
* Adjusting the way our processing loop works to permit multiple commands per cycle (the same goes for the adapters?)

With everything else going on, this is quite low priority for now, unless it becomes an issue in real usage, so putting it in the backlog.

MichaelWedel:
This fixes #143.

A method `set_device_parameters` has been added to `Simulation` as per the discussion in #143. Documentation and release notes have been updated.

### DIFF
--- a/docs/release_notes/release_1_0_2.rst
+++ b/docs/release_notes/release_1_0_2.rst
@@ -56,7 +56,7 @@ New features
 
    ::
 
-      $ lewis-control device set_parameters "{'target_speed': 1, 'target_phase': 20}"
+      $ lewis-control simulation set_device_parameters "{'target_speed': 1, 'target_phase': 20}"
 
    Thanks to the IBEX team for requesting this.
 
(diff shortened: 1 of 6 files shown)


MikeHart85:
Looks good, merging!

PR #185, opened 2017-01-26
Added command line tools help output to generated documentation

### LINKED ISSUE #176: Add documentation for command line scripts
Currently we do not have auto-generated documentation for command line script usage. There is a Sphinx extension (http://sphinx-argparse.readthedocs.io/en/latest/index.html) that extracts that from the argparser of a script.

This should be added to the user guide, while the documentation should also be updated a bit to be a bit nicer and more descriptive where it isn't.

MichaelWedel:
This fixes #176.

After trying different extensions to do automatic documentation of ArgumentParsers in sphinx, I decided to solve the problem without any extensions and just use the module docstring. I think if https://pythonhosted.org/sphinxcontrib-autoprogram/ has advanced one or two releases we could switch to that, as it produces quite nice lists, but 0.1.3 only works with Python 3 (which will be fixed in the next release) so at the moment it's not an option.

I also made the help a bit more detailed and tried putting the arguments into more logical categories than "positional" and "optional".

### DIFF
--- a/docs/user_guide/command_line_tools.rst
+++ b/docs/user_guide/command_line_tools.rst
@@ -1,0 +1,15 @@
+Command line tools
+==================
+
+This page documents the program usage for ``lewis`` and ``lewis-control``, the two command line
+tools provided as part of a Lewis installation.
+
+lewis
+-----
+
+.. automodule:: lewis.scripts.run
+
+lewis-control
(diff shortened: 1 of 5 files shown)


MikeHart85:
Looking good, merging!

PR #186, opened 2017-01-26
Update copyright notices

### LINKED ISSUE #179: Update Copyright Notices for 2017
As noted in #175, copyright notices should be updated, to read:

> Copyright (C) 2016-2017 European Spallation Source ERIC

This is technically optional; we could leave it as 2016 and continue using 2016 even in new files. 

But if we do want to update, we should do it wholesale in all files across the project.

See also: https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

MichaelWedel:
This fixes #179.

Nothing to say except...it updates all copyright notices!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,7 +23,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'lewis'
-copyright = u'2016, European Spallation Source ERIC'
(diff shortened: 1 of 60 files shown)


MikeHart85:
Looks good, merging!

PR #192, opened 2017-02-07
Fix control server/client issues

### LINKED ISSUE #191: Control server/client issues
There are a few issues with the control server and client:

- Objects can't be removed on the server side, this is a prerequisite for #98.
- When exposing properties they are actually obtained using `getattr`, which might have side effects.
- The `ControlClient` must implement some timeout if a connection can't be established or a request can't be completed.

These are closely related and can be handled as one issue.

MichaelWedel:
This fixes #191.

It is now possible to remove exposed objects, although this is not used yet. But it will be important when we implement setup-switching (where the "old" device must be removed and the "new" device must be exposed). Properties are no longer called when the wrapper methods on the `ControlServer` are created. The connection of `ControlClient` times out after a user specified amount of time, the default is 1 second. Tests for the new behavior have been added, documentation (including release notes) has been updated.

### DIFF
--- a/lewis/core/control_client.py
+++ b/lewis/core/control_client.py
@@ -26,7 +26,7 @@ This module provides client code for objects exposed via JSON-RPC over ZMQ.
     in the module :mod:`~lewis.core.control_server`.
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import zmq
 import uuid
@@ -73,16 +73,30 @@ class ControlClient(object):
     of objects at the top level, a dictionary of named objects can be
     obtained via get_object_collection.
 
-    :param host: Host the control server is running on
-    :param port: Port on which the control server is listening
+    If a timeout is supplied, all underlying network operations time out
+    after the specified time (in milliseconds). The default is 1000 ms,
+    for no timeout specify ``None``.
+
+    :param host: Host the control server is running on.
+    :param port: Port on which the control server is listening.
+    :param timeout: Timeout in milliseconds for ZMQ operations.
     """
 
-    def __init__(self, host='127.0.0.1', port='10000'):
+    def __init__(self, host='127.0.0.1', port='10000', timeout=1000):
+        self.timeout = timeout if timeout is not None else -1
+
         self._socket = self._get_zmq_req_socket()
-        self._socket.connect('tcp://{0}:{1}'.format(host, port))
+
+        self._connection_string = 'tcp://{0}:{1}'.format(host, port)
+        self._socket.connect(self._connection_string)
 
     def _get_zmq_req_socket(self):
         context = zmq.Context()
+        context.setsockopt(zmq.REQ_CORRELATE, 1)
+        context.setsockopt(zmq.REQ_RELAXED, 1)
+        context.setsockopt(zmq.SNDTIMEO, self.timeout)
+        context.setsockopt(zmq.RCVTIMEO, self.timeout)
+        context.setsockopt(zmq.LINGER, 0)
         return context.socket(zmq.REQ)
 
     def json_rpc(self, method, *args):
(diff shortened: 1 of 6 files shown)


lewis/core/control_client.py
```diff
@@ -73,16 +73,30 @@ class ControlClient(object):
...
+    after the specified time (in milliseconds). The default is 1000 ms,
+    for no timeout specify ``None``.
+
+    :param host: Host the control server is running on.
+    :param port: Port on which the control server is listening.
+    :param timeout: Timeout in milliseconds for ZMQ operations.
     """
 
-    def __init__(self, host='127.0.0.1', port='10000'):
+    def __init__(self, host='127.0.0.1', port='10000', timeout=1000):
```
MikeHart85: At a glance, 1 second seems a bit short. Maybe something like 3 seconds would work better to avoid timing out on brief network lag?

The comment above would need to change as well.

lewis/core/control_client.py
```diff
@@ -98,14 +112,20 @@ def json_rpc(self, method, *args):
...
+        try:
+            self._socket.send_json(
+                {'method': method,
+                 'params': args,
+                 'jsonrpc': '2.0',
+                 'id': request_id
+                 })
+
+            return self._socket.recv_json(), request_id
+        except zmq.error.Again:
```
MikeHart85, line 124: If the device disconnects **during** listing properties of a `lewis-control device`, this will still go through each property one-by-one and wait for each one to time out individually. With devices that have many properties, this can quickly add up to a lot of time. Especially if we increase the timeout duration.

Generally, this is a symptom  of only one parameter being able to be read per cycle. Perhaps the best solution would be a `get_device_parameters` equivalent for the `set_device_parameters` introduced in #182?

To test, run tsdauwrapper (a relatively slow cycle rate helps), enter `lewis-control device` into a console, and immediately close the tsdau dialog.

lewis/core/control_client.py
MikeHart85: Actually, the real default comes from the argparser in `run_simulation`, of course... which raises the question of whether there's any value in having a default here at all? (I think we wondered the same about host and port at one point?)

lewis/core/control_server.py
```diff
@@ -191,6 +194,22 @@ def add_object(self, obj, name):
             glue = '.' if not method_name.startswith(':') else ''
             self._add_function(name + glue + method_name, exposed_object[method_name])
 
+    def remove_object(self, name):
```
MikeHart85, line 197: What is the purpose / intended usage of this function? Not seeing any usage outside of tests.

lewis/scripts/control.py
```diff
@@ -108,6 +108,10 @@ def call_method(remote, object_name, method, arguments):
     '-r', '--rpc-host', default='127.0.0.1:10000',
     help='HOST:PORT string specifying control server to connect to.')
 optional_args.add_argument(
+    '-t', '--timeout', default='1000', type=int,
```
MikeHart85: Would need to change here as well if you decide to change the timeout.

lewis/scripts/control.py
```diff
@@ -122,15 +126,19 @@ def call_method(remote, object_name, method, arguments):
...
-                print(response)
+            if not args.member:
+                show_api(remote, args.object)
+            else:
+                response = call_method(remote, args.object, args.member, args.arguments)
+
+                if response is not None or args.print_none:
+                    print(response)
+    except ProtocolException as e:
+        print('\n'.join(('An error occured:', str(e))))
```
MikeHart85: Typo in "occurred".

lewis/core/control_client.py
MichaelWedel: I think it's okay to have a default, we could make it longer. 1 second was just some arbitrary number...like 3. I'm happy to change it to 3.

lewis/core/control_client.py
MichaelWedel: I also tried that, but with one of the inbuilt lewis devices, I did not see that behavior. But I'm happy to change to implementation of `lewis-control` for this.

lewis/core/control_server.py
MichaelWedel: As I wrote in the PR notes above, we're going to need it for other tickets when we start switching out objects in the simulation.

lewis/scripts/control.py
MichaelWedel: My usual typo, thanks. Found some in the release notes as well.

lewis/core/control_server.py
MikeHart85: Ah, it helps to read the PR description :laughing: Makes sense now :+1:

MikeHart85:
Looks good :+1:

PR #196, opened 2017-02-09
Truncate multiline string properties in control.py

MichaelWedel:
This has no issue, but it is a fairly annoying bug. This change truncates multiline message and indicates further content by appending `[...]`.

### DIFF
--- a/lewis/scripts/control.py
+++ b/lewis/scripts/control.py
@@ -48,7 +48,8 @@ def show_api(remote, object_name):
             raw_value = str(getattr(obj, prop))
             value_lines = raw_value.split('\n')
 
-            current_value = value_lines[0][40:] + (' [...]' if len(value_lines) > 1 else '')
+            current_value = value_lines[0][40:] + (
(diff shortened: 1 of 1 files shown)


lewis/scripts/control.py
```diff
@@ -45,7 +45,10 @@ def show_api(remote, object_name):
     maxlen = len(max(properties, key=len))
     for prop in sorted(properties):
         try:
-            current_value = getattr(obj, prop)
+            raw_value = str(getattr(obj, prop))
+            value_lines = raw_value.split('\n')
+
+            current_value = value_lines[0] + (' [...]' if len(value_lines) > 1 else '')
```
MikeHart85: Consider something like (untested):
```python
current_value = value_lines[0][:20] + (' [...]' if len(value_lines) > 1 or len(value_lines[0]) > 20 else '')
```
Could have really long lines without newlines at some point.

MichaelWedel: Good point, although 20 seems a bit short...maybe 40? So in total we're still below the magical command line 80 :)

PR #197, opened 2017-02-10
Expose device interface separately

### LINKED ISSUE #173: Expose device interface separately in ControlServer
As discussed in and a prerequisite for #171, the device interface needs to be handled differently in `Simulation`.

There needs to be an intermediate object that owns the adapters and is exposed via `ControlServer`. This is also a preparation step for allowing multiple interfaces to one simulated device.

MichaelWedel:
This fixes #173.

Several things have changed:

- `Adapter` was moved to `lewis.core.adapters`, equivalent to `lewis.core.devices`
- In this new module is a new class `AdapterContainer` that manages a collection of adapters. This is used in `Simulation` now instead of storing adapters directly. The `AdapterContainer` is exposed as `interface` via the control server. The `connect_device` and so methods from `Simulation` have been removed.
- Unit tests were added, also for `ForwardProperty` which had to be moved to `lewis.core.utils` to make `lewis.adapters` empty on the top level.

The interface of `AdapterContainer` is flexible, the relevant functions accept an arbitrary number of protocol names and just address all adapters if none were provided (this was discussed in #173 or related issue). So now you can do:

```
$ ./lewis.py -r localhost:10000 -k lewis.examples example_motor
```

And then a new world of commands is available:

```
$ ./lewis-control.py interface is_connected
{'stream': True}
$ ./lewis-control.py interface is_connected stream
True
$ ./lewis-control.py interface disconnect
$ ./lewis-control.py interface connect stream
````

Since all these functions use `*args`, this prepares for multiple adapters.

What's still missing now is end user documentation (wanted to make sure everything else is agreed upon first). What we need to change is the implementation of `handle`, because this only works correctly with one adapter now. But I would suggest we push that into the future when we actually have multiple adapters - now it behaves exactly as before for all cases that we support in `Simulation`.


MichaelWedel:
The codacy error is not accurate, the property is implemented...

Issue #199, opened 2017-02-13
Introduce device "sanity check"


MichaelWedel:
This is somewhat related to #198 and also #126.

There are certain things that `lewis` expects, such as that there's a device and an adapter in a device module (this is subject to change, but there will be other requirements).

While we can't make sure that user-written devices behave correctly, we could make it easier to develop them by providing a testing script which verifies that a device corresponds to what `lewis` expects. What I imagine is a script that takes a device and tries to spot issues:

```shell
$ ./lewis-verify-device -k my.own.module new_device
Warning: Device uses deprecated method x, consider switching to y for future-proofing.
Warning: Module does not provide an adapter, are you sure that this is intended?
Warning: Device attribute x is not JSON serializable, there might be problems with the control server.
```

It would also provide a way for users to verify that devices from a different source are compatible with the installed Lewis version.

PR #200, opened 2017-02-13
Devices can specify required framework version(s)

### LINKED ISSUE #198: Mechanism for devices to indicate dependency on certain framework version
As we have discussed in various contexts, since we support Lewis being used as a library with people writing their own device simulators that are in places we have no control over, we should make sure stuff does not break horribly as we advance the framework.

`DeviceBase` should have an additional attribute or property that can be overridden by device-implementers to indicate the framework version(s) the device is known to work with. This should be similar to what `pip` offers, so that users can specify things like `>=1.0.2` or `==1.0.0`. By default, nothing can be assumed or verified, so this is something that we have to point out specifically in the documentation.

The device discovery mechanism should still detect devices that are incompatible with the framework, but it should store them separately so that `lewis` can list them as "incompatible devices".

MichaelWedel:
This fixes #198.

With these changes it is possible to specify a module-level variable `framework_version` which lets a device author specify which version of Lewis the device has been verified to work with. Examples are `1.0.3` or more complex specifications such as `>=1.0.3, <1.0.6`. For each device module that is not compatible with the framework, `lewis` logs an error message. This will help users diagnose why a device is not available if loaded from an external module (via `-k`). If the variable is not specified the device will appear in all versions of Lewis, but the documentation recommends not to do this. The reason is simplicity, just like we're not enforcing the presence of `setups`.

The relevant documentation as well as release notes have been updated.

To test, you could modify one of the example devices to require a higher (or lower) framework version.


MikeHart85:
```python
framework_version = '>=1.0.3,<2.0.0'
```
This is a really cool way to specify versions. I particularly like how easy it is to do with that module. But after some testing and thinking about it for a while, I wonder if this is a reasonable approach to versioning our devices.

A device cannot really know which framework versions it is compatible with, particularly for future versions. Even for older versions, the responsibility for knowing (in)compatible framework versions should not be with the device developer.

All a device can / should need to know is which single framework version it was developed for. Only the framework itself can / should know whether it (at its current version) can support a device written for a given version. Newer versions are still a big question mark, of course... all you can really do is issue an error or warning (I think version mismatch should always be just a warning, since everything might work fine... you just don't know).

The framework could then also use the specific `framework_version` of a device to run in "compatibility mode" so to speak. If we make a breaking change to how something works, we might be able to leave legacy support for it behind a conditional check.

The `framework_version` then also doesn't need to / should not change with every new framework release. It should only be updated when the device is updated, and even then only if it uses a new feature that the old framework version did not support.

MichaelWedel:
Thanks for your comment! I also thought about this somewhat, but I think this must also be seen in the "grand scheme of things" with device verification coming up and so on.

My reasoning for doing it this way (for now) was the following. The device developer knows how the device should behave, so they are the only one who can verify that "everything is fine". Once they've tested against a new version of the framework, the `framework_version` member can be modified to make it "official". If it is not compatible, or the author can not or does not want to update the device, the user still has the option of installing an older version of Lewis in a different virtual environment. Short of knowing whether a device works or not, I simply thought this was the safest way of going about this.

So far for "justification" why I did what I did. I still think version ranges and so on are a valid use case for authors that take good care of their devices and verify it against new framework versions - with proper tests in place (now we're talking about the future) that should be easy most of the time. And then I don't see a reason why a developer shouldn't be allowed to say "this works with everything between 1.0.2 and 3.4.5 but then it stops because feature x was changed".

What I could imagine is a sort of "ignore version information"-flag where the devices are still loaded (with a big fat warning though) and it's possible to run them "without any guarantees", that would be easy to implement and it would give users the power to decide for themselves.

Regarding compatibility mode, I like the idea, but it's definitely out of scope for this ticket (and likely even this release). We should open an issue and keep an eye on it.

What do you think, would it be enough for now to add the flag to `lewis` and leave it as it is now? As I wrote, I expect some interplay and changes with device verification, system tests and so on.

MichaelWedel:
After our discussion I removed the possibility to provide proper version specs (although I have a suspicion they will come back at some point - as discussed, easier to add them back in than to take them out).

Devices are loaded regardless of the framework version they specify, but when there is a mismatch a warning is logged and the user is presented with an error message that contains a hint to the new `-R/--relaxed-versions`-flag. With that flag the error message goes away and the device is usable, but the warning will still be there. I think that's a fair compromise between letting users know that some things might not work and not absolutely blocking them if they are willing to take the risk (or they have tested and it works).

The warning message and the details of the version verification are likely to change as part of #199.

lewis/core/devices.py
```diff
@@ -59,6 +60,16 @@ class DeviceBuilder(object):
     :class:`DeviceRegistry`) and inspects it so that it's possible to construct devices and
     interfaces.
 
+    Before loading anything, version compatibility between the device module and the framework is
+    verified. If the module contains a ``framework_version`` variable at the top level holding a
+    version spec (for example ``framework_version='>=1.0.0'`` or
+    ``framework_version='>=1.0.1,<1.0.3'``), this spec is checked against the current version of
```
MikeHart85: Some left over references to version ranges here.

lewis/core/devices.py
```diff
@@ -327,17 +343,44 @@ def devices(self):
...
+    def device_builder(self, name, relaxed_versions=False):
         """
         Returns a :class:`DeviceBuilder` instance that can be used to create device objects
         based on setups, as well as device interfaces. If the device name is not stored
         in the internal map, a LewisException is raised.
 
+        Each DeviceBuilder has a ``framework_version``-member, which specifies the version
+        of Lewis the device has been written for. If the version does not match the current
+        framework version, it is only possible to obtain those device builders calling the
+        method with ``strict_versions`` set to ``False``, otherwise a
```
MikeHart85: Shouldn't this be "`relaxed_version` set to `True`"?

lewis/devices/chopper/__init__.py
```diff
@@ -19,4 +19,6 @@
 
 from .device import SimulatedChopper
 
+framework_version = '1.0.3'
```
MikeHart85: This should probably be reduced to `1.0.2` for now, like the others.

lewis/examples/example_motor/__init__.py
```diff
@@ -146,3 +146,5 @@ def set_target(self, new_target):
         )
     )
 )
+
+framework_version = '>=1.0.3'
```
MikeHart85: This should probably be reduced to `1.0.2` for now, like the others, and the ranging removed.

lewis/examples/modbus_device/__init__.py
```diff
@@ -40,3 +40,6 @@ class ModbusInterface(ModbusAdapter):
     co = di
     ir = ModbusBasicDataBank(0)
     hr = ir
+
+
+framework_version = '>=1.0.3'
```
MikeHart85: This should probably be reduced to `1.0.2` for now, like the others, and the ranging removed.

lewis/examples/simple_device/__init__.py
```diff
@@ -67,3 +67,6 @@ def set_param(self, new_param):
 
     def handle_error(self, request, error):
         return 'An error occurred: ' + repr(error)
+
+
+framework_version = '>=1.0.3'
```
MikeHart85: This should probably be reduced to `1.0.2` for now, like the others, and the ranging removed.

MikeHart85 (commented):
Just some minor tweaks of stuff left over from before.

PR #203, opened 2017-02-20
Refactor binding step in EpicsAdapter

### LINKED ISSUE #202: EpicsAdapter should follow binding concept
In `StreamAdapter` we switched from modifying the Adapter at runtime to map Device methods to a "binding approach". This approach made things much easier and made `ForwardMethod` disappear which wasn't that useful in other contexts.

By moving to a similar approach it will be possible to get rid of `ForwardMethod` as well (which is even harder to use in other contexts). Furthermore, the two Adapters get more similar again, maybe we see some more possibilities for convergence later on.

Functionally this will not change anything.

MichaelWedel:
This fixes #202.

Rather than modifying `EpicsAdapter` when binding an interface to a device, `EpicsAdapter` now uses a new helper class `BoundPV`, which is somewhat similar to `Func` in `StreamAdapter`.

As a consequence, `ForwardProperty` is not required anymore, so it has been removed. Updates to documentation are not required, as this was a purely internal change.


lewis/adapters/epics.py
```diff
@@ -95,6 +94,76 @@ def __init__(self, target_property, poll_interval=1.0, read_only=False,
...
+
+    @property
+    def value(self):
+        """Value of the bound property on the target."""
+        return getattr(self._target, self._pv.property)
+
+    @value.setter
+    def value(self, new_value):
+        if not self.read_only:
+            setattr(self._target, self._pv.property, new_value)
```
MikeHart85: Would something like this be safer?
```python
if self.read_only:
    raise ...
setattr(...)
```
Otherwise attempts to write read only values are ignored silently.

lewis/adapters/epics.py
```diff
@@ -145,8 +212,9 @@ def process_pv_updates(self, force=False):
 
         self.updatePVs()
```
MikeHart85, line 216: Do you remember why we silently ignore AttributeError and TypeError just above here?

lewis/adapters/epics.py
```diff
@@ -95,6 +94,76 @@ def __init__(self, target_property, poll_interval=1.0, read_only=False,
...
+
+    @property
+    def config(self):
+        """Config dict passed on to pcaspy-machinery."""
+        return self._pv.config
+
+    @property
+    def poll_interval(self):
+        """Interval at which to update PV in pcaspy."""
+        return self._pv.poll_interval
```
MikeHart85, line 161: There are a few properties that really just pass through to `self._pv`. For now it's manageable, but if this grows it might be good to consider using `__getattr__` to pass through in a generic way. The downside would be that intellisense probably wouldn't work as well.

lewis/adapters/epics.py
```diff
@@ -218,26 +286,46 @@ def __init__(self, device, arguments=None):
...
 
         self._options = self._parse_arguments(arguments or [])
-
-        self._create_properties(self.pvs.values())
+        self._bound_pvs = self._bind_properties(self.pvs)
 
         self._server = None
         self._driver = None
 
+    def _bind_properties(self, pvs):
```
MikeHart85, line 297: This would benefit from a docstring, particularly about the format of `pvs` and the return value.

MikeHart85 (commented):
Looks great, just some minor comments / questions.

lewis/adapters/epics.py
MichaelWedel: It might have been a "safety net" from very early days when there was a risk that the property did not exist on the adapter. Another thing that can go wrong is that the property getter returns an invalid value type, so that `setParam` might raise one of these exceptions - but I'm not entirely sure about it.

lewis/adapters/epics.py
MichaelWedel: :+1:

lewis/adapters/epics.py
MichaelWedel: Yes we should keep an eye on that. It might also happen that `BoundPV` and `PV` diverge, but it's a bit hard to say.

lewis/adapters/epics.py
MichaelWedel: It might be worthwhile, but essentially in the context of our `Driver` this can never happen because it's caught before:

https://github.com/DMSC-Instrument-Data/lewis/pull/203/files#diff-0221a74d743968a9172db2eaefb8b4a2R184

As in the comment below, this might be something from a simpler time :)

MikeHart85:
Looks good, merging!

Issue #204, opened 2017-02-21
Maintain devices in separate package


MichaelWedel:
We've discussed this earlier and after thinking about it for a while I believe it's a good idea to separate the devices into their own package that can be installed and maintained.

This consequently drives Lewis more in the direction of a library rather than a simulator for a few chosen devices.

There are some things that need more thought beforehand, for example unit tests. I think the unit tests for devices should be part of the device package like so:

```
devices/
  device_1/
    interfaces/
    other_stuff/
    tests/
  device_2/
    interfaces/
    tests/
```

Running these tests could be part of the "device verification" suggested in #199.

MichaelWedel:
I did not want to put it into 1.0.3 just yet, there a quite a few unanswered questions or other issues that need some work first. So it's in the backlog for now with the option of moving it to the release should it be realistic.

MikeHart85:
Perhaps it would make sense to restructure devices so the top folder is empty (or perhaps just contains a config file or the like):
```
/
  device_1/
    device/
    interfaces/
    setups/
    tests/
  device_2/
    device/
    interfaces/
    setups/
    tests/
```

MichaelWedel:
We just have to keep the packaging in mind and it also depends a bit how much "stuff" there is at the top level. If there are docker files, travis config and so on I'd probably prefer doing it like it is with Lewis now (with a directory).

We'll see how it develops.

MikeHart85:
No, I meant the top level inside each `device_1`, `device_2`, etc. Not the top top level. Sorry, confusing wording.

MichaelWedel:
Ah okay, now I see. An empty `__init__.py` file might still be required so that the import stuff works properly.

PR #205, opened 2017-02-21
Improve device discovery

MikeHart85:
The need for importing and `__all__`ing device classes in `__init__.py` has been a thorn in my eye for a while. Seeing it again (and being expanded) in #200 was the last drop.

Searching `device.py` (in addition to the top level device module itself) for device types seems to remove the need for doing this. This way, the `__init__.py` stays clean and doesn't need to be updated whenever device classes are added/removed/renamed.

One thing that still bugs me with this approach is that it makes `device.py` a special filename. This seems to be the de facto standard anyway (all devices I've seen use this filename). But I still don't like it. Perhaps a future change to think about (hinted in #204) would be to have a special directory called `device` or `devices`. Then all `.py` files inside that directory can be scanned for devices, no matter what they are called. This would be consistent with how interfaces work.

I've also removed all the `import`/`__all__`s in the `interfaces/__init__.py` files. No change in behaviour was necessary to make this work... interface discovery was already smart enough to find the interfaces without them; it seems they were redundant already.

An unused / commented out chunk of code was also removed.


lewis/core/devices.py
```diff
@@ -112,16 +112,9 @@ def __init__(self, module):
...
-            # syslog.warning('No devices discovered in module {}.'.format(module.__name__))
-            pass
-        else:
-            pass
-            # syslog.info('Discovered devices in module {}: {}'.format(
-            #    module.__name__, ','.join([x.__name__ for x in self._device_types])))
-
         submodules = get_submodules(self._module)
 
+        self._device_types += list(get_members(submodules.get('device'), is_device).values())
```
MichaelWedel: For consistency it might be nice to do it like for the interfaces and setups (like you suggested in your comment) and look around a bit (in case it's not a single `device.py` but a `device/`-directory with multiple files in it). So basically very similar to how `_discover_interfaces` works, but `_discover_device_types`. In fact the method would be more or less the same (even line 113 would go away). Much of the code could be shared I suppose, it's mostly `is_device` vs `is_adapter`.

MichaelWedel (commented):
Thanks, I was also quite annoyed by that in #173, so it's good to fix it. You already pointed out that for consistency it would be better to do it like `interfaces` so I think this could be a part of this PR, then it's done properly, please see the other comment as well.

lewis/core/devices.py
MikeHart85: Do you think `states.py` and (for the chopper) `bearings.py` should also go into the `device/` directory?

That would be my gut instinct. Then we have a separation between behaviour (everything in `device/`) and interface (everything in `interfaces/`). Also it mirrors the control interface nicely.

`device/` or `devices/` (to match `interfaces/`, `setups/`, `tests/`, ...)?

MichaelWedel: I think our gut instincts agree, and I'd also say `devices`, because there might be more than one device type.

MikeHart85:
This should hopefully do what we discussed... turned out much simpler than `_discover_interfaces` since it just returns a list of any device type it finds.

lewis/core/devices.py
```diff
@@ -111,21 +111,25 @@ class DeviceBuilder(object):
...
-        else:
-            pass
-            # syslog.info('Discovered devices in module {}: {}'.format(
-            #    module.__name__, ','.join([x.__name__ for x in self._device_types])))
-
         submodules = get_submodules(self._module)
 
+        # Devices can be in __init__.py, device.py, or any devices/*.py
+        self._device_types = list(get_members(self._module, is_device).values())
+        self._device_types += list(get_members(submodules.get('device'), is_device).values())
```
MichaelWedel: I think either these two cases should be in `_discover_devices`, like for the interfaces and setups. That would actually make all of these discovery methods consistent.

Looking at the `device` case and the way you shuffled around the files for the current devices I'd even raise the question whether we really need that case, it just adds another "magic name". For simple cases I think providing devices in the  `__init__.py` (which is mandatory as a file anyway) is enough, all other cases can go into `devices`. Plus, we're not breaking anything by not adding this possibility.

test/test_SimulatedChopper.py
```diff
@@ -19,8 +19,8 @@
 
 import unittest
 
-from lewis.devices.chopper.states import DefaultIdleState
-from lewis.devices.chopper import SimulatedChopper
+from lewis.devices.chopper.devices.states import DefaultIdleState
+from lewis.devices.chopper.devices.device import SimulatedChopper
```
MichaelWedel, line 23: This looks a bit weird to me. Is there a better filename than `device`? It seems the word `device` is repeated a lot of times now.

lewis/core/devices.py
MikeHart85: You're right, `device.py` can be removed. I left it in because I was thinking of the devices that the IBEX team created (they have a `device.py`). But they should not break anyway because they still have the `__all__` stuff in their `__init__.py`. It just means they have to put their `device.py` into `devices/` if they want to get rid of the `__all__` stuff.

As for putting the additional cases (now only `__init__.py`) into `_discover_devices`... I'm a bit conflicted. I had the same thought as you initially, but decided to leave them separate because I felt that it would be *in*consistent with the other two discovery methods. As they are, they all have a signature like this:

```python
_discover_something(some_module)
```
and return whatever `something`s they found in `some_module`.

If we move `__init__.py` discovery into `_discover_devices`, it will now return whatever it found in `some_module` _plus_ whatever it found in `__init__.py`. If, at some point in the future, we want to also load devices from `some_other_module`, we won't be able to just add one line like this:

```python
self._device_types += self._discover_devices(submodules.get('devices'))
self._device_types += self._discover_devices(submodules.get('some_other_module'))
```
because now we'll have anything found in `__init__.py` returned twice.

test/test_SimulatedChopper.py
MikeHart85: It does look weird now, but I'm not sure how to resolve it. What should we call it instead?

If we call it `chopper.py`, we just wind up with:
```python
from lewis.devices.chopper.devices.chopper import ...
```
Which isn't exactly an improvement.

Keep in mind though this is a structure change in preparation for more changes to come (which may fix this awkward naming):
- Devices should soon be moved to their own repository (where we hopefully won't have a top-level `devices/` anymore)
- Tests should soon be moved to a `tests/` directory alongside `devices/` and `interfaces/`

I think the import statements will become clearer as we make those changes.

lewis/core/devices.py
MichaelWedel: The one for `interfaces` does exactly this. It discovers whatever is `__init__.py` plus whatever is in the provided module.

MichaelWedel: I can see you argument, why not make both methods accept `*modules` and include `self._module` in the argument list.

MikeHart85: It wouldn't work that simply because `__init__.py` is a single module and (I just realize we've used incorrect terminology here and in the code) `devices/`, `interfaces/`, etc are packages.

So `discover_something` actually takes a `some_package` parameter, and iterates over the modules in that package to find classes that fit `something`. This wouldn't work with passing in a module.

I don't know, I'll tinker a little more with it... I'm sure there's a way to make this more consistent.

MichaelWedel: That is actually true, it's slightly different. But yes, there must be a way to make it consistent (probably other ways than how it's implemented in the interface discovery).

MikeHart85:
Ended up just removing `device.py` support and moving `__init__.py` support into `_discover_devices()`. It's not ideal, but it should work for now.

PR #206, opened 2017-02-24
Add unit tests for NotImplementedErrors

MichaelWedel:
This has been bothering me for some time, these tests check that the `NotImplementedErrors` in the core framework are actually raised.

### DIFF
--- a/test/test_StateMachineDevice.py
+++ b/test/test_StateMachineDevice.py
@@ -19,7 +19,7 @@
 
 import unittest
 
-from mock import Mock, call
+from mock import Mock, call, patch
 
 from lewis.devices import StateMachineDevice
 from . import assertRaisesNothing
@@ -35,6 +35,17 @@ class MockStateMachineDevice(StateMachineDevice):
 
 
 class TestStateMachineDevice(unittest.TestCase):
+    def test_not_implemented_errors(self):
+        self.assertRaises(NotImplementedError, StateMachineDevice)
+
+        with patch('lewis.devices.StateMachineDevice._get_state_handlers',
+                   return_value={'test': Mock()}):
+            self.assertRaises(NotImplementedError, StateMachineDevice)
+
+            with patch('lewis.devices.StateMachineDevice._get_initial_state',
+                       return_value='test'):
+                self.assertRaises(NotImplementedError, StateMachineDevice)
+
     def test_init_calls_appropriate_methods(self):
         smd = MockStateMachineDevice()
 
--- a/test/test_StateMachineDevice.py
+++ b/test/test_StateMachineDevice.py
@@ -18,8 +18,9 @@
 # *********************************************************************
 
 import unittest
+import itertools
 
-from mock import Mock, call, patch
+from mock import Mock, MagicMock, call, patch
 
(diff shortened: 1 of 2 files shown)


test/test_StateMachineDevice.py
```diff
@@ -35,6 +35,17 @@ class MockStateMachineDevice(StateMachineDevice):
...
+    def test_not_implemented_errors(self):
+        self.assertRaises(NotImplementedError, StateMachineDevice)
+
+        with patch('lewis.devices.StateMachineDevice._get_state_handlers',
+                   return_value={'test': Mock()}):
+            self.assertRaises(NotImplementedError, StateMachineDevice)
+
+            with patch('lewis.devices.StateMachineDevice._get_initial_state',
+                       return_value='test'):
+                self.assertRaises(NotImplementedError, StateMachineDevice)
```
MikeHart85: This test will pass if only `_get_transition_handlers` raises that exception, and `_get_state_handlers` and `_get_initial_state` do nothing (or something else).

It will also pass if there is some other base method that also raises `NotImplementedException` on construction. Perhaps this is fine like that, just pointing it out in case it's not what you intended.

MichaelWedel: The first case you describe is the one I had in mind, this is why the methods are patched (so that they don't raise the exception before the other ones have a chance to do that) - I think that does what I wanted it to do.

You're right about the second case, this might happen if one of the bases of `StateMachineDevice` (`DeviceBase` and `CanProcessComposite`) decide to introduce methods that need to be implemented. A change like that basically changes the interface of the class, so the unit tests would likely require change as well.

Is there a better way to test this?

MikeHart85: I guess it's a question of what exactly the test is trying to guarantee about framework behaviour. My guess (which may be wrong) is that you're trying to prove/ensure that...

> This base class has methods A, B and C that must be implemented by derived classes. 
>
> Construction fails if __any__ of A, B and C are not implemented. 
>
> Construction succeeds if __all__ of A, B and C are implemented.

In the first case my complaint is that you're not really testing that construction fails if __any__ of A, B or C are not implemented. Because the test will still pass if A and B do not throw such exceptions. Actually, it will also pass if C doesn't throw such an exception, but some other method X that is called during construction does.

In the second case I'm saying there is no test for successful construction when __all__ of A, B and C are implemented.

To test the conditions I quoted, you'd need to do something like:

```python
def test(...):
    class CaseAMissing(BaseClass):
        B = None
        C = None
    self.assertRaises(NotImplementedError, CaseAMissing)

    class CaseBMissing(BaseClass):
        A = None
        C = None
    # ...
    # ...
    class CaseAllImplemented(BaseClass):
        # A, B, C = None
    assertRaisesNothing(self, CaseAllImplemented)
```
Although that's a very brute force approach... maybe there's a cleaner way of implementing that.

MichaelWedel:
This is the shortest way I was able to come up with to test each missing method and an "implementation" with all methods present.

MikeHart85:
That's a cool way of solving it, I like it. Ready to merge?

PR #208, opened 2017-03-01
Clean up adapter documentation

MichaelWedel:
After moving adapters to `core.adapters`, we forgot to also modify the documentation. This fixes the documentation, plus the core APIs are now alphabetically ordered which is easier to scan through.

### DIFF
--- a/docs/api/adapters.rst
+++ b/docs/api/adapters.rst
@@ -1,14 +1,11 @@
 Adapter API
 ===========
 
-The Adapter API consists of a general :mod:`~lewis.adapters`-module that defines a base class
-for all specific adapters. These specific adapters in turn define slightly different utilities
-for writing device interfaces.
+.. automodule:: lewis.adapters
 
 .. toctree::
     :maxdepth: 2
 
-    adapters/adapters
     adapters/epics
     adapters/modbus
     adapters/stream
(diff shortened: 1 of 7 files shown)


lewis/adapters/__init__.py
```diff
@@ -18,7 +18,7 @@
 # *********************************************************************
 
 """
-The core Adapter API is located in :mod:`lewis.core.adapters`. This module on the contains
+The core Adapter API is located in :mod:`lewis.core.adapters`. This package on the contains
```
MichaelWedel: :+1:

docs/api/adapters.rst
```diff
@@ -1,14 +1,11 @@
...
-for writing device interfaces.
+.. automodule:: lewis.adapters
 
 .. toctree::
     :maxdepth: 2
 
-    adapters/adapters
     adapters/epics
     adapters/modbus
     adapters/stream
```
MikeHart85, line 11: The titles this generates (based on the rst files in `docs/api/adapters/`) are:
```
    The EPICS-Adapter
    The Modbus Adapter
    The Stream-Adapter
```
The dashes are inconsistent. I think no dash makes more sense?

MikeHart85 (commented):
Just the one comment about dashes. Not sure if you agree?

I also went ahead and changed "module" to "package" in `adapters/__init__.py`.

docs/api/adapters.rst
MichaelWedel: Yes, seems better without dashes at all, while we're at it.

Issue #98, opened 2016-09-20
Ability to change setups at runtime


MikeHart85:
One aspect of #87 that was not addressed in #93 is switching setups at runtime.

Now that we have a way to control aspects of the simulation at runtime, it should be possible to load different setups of the same device at runtime as well. 

This would enable testing different failure scenarios without restarting the simulation. It could also be used to put a device into a mode where it just echoes the values you set to attributes, rather than actually running a real simulation (this could be done using a setup that forces it to stay in a neutral state that doesn't have any behaviour). This would also be useful for testing purposes.

This should be easier to do once creation of adapter and device instances is moved to the central `Simulation` class.

One technical challenge of this might be ensuring the adapter stays up, rather than shutting down and coming back online (if a new instance needs to be created).

MichaelWedel:
Just as a question, is it problematic to restart a simulation to test a different scenario? If the concern is that it would take too long or it would be too repetitive to get the simulation back to the state it was in, maybe addressing that could be another way.

I'm basically thinking of taking a snapshot of the device at a certain point in time (all the state, including current state of the statemachine) and applying that again, possibly to a device implementation with different state handlers in the state machine (for failure cases).

Regarding echo-mode, I somehow suspect that there might be other ways (tools) to achieve this. I'm not entirely sure how this "not simulating the device" fits with with our (current) goal of providing a device simulation. Maybe this is something for the long term.

MichaelWedel:
I had another thought regarding "echo-mode". Might this be solved at least partly be the logging improvements we've been discussing? Ideally, the Adapter would log at some level what commands it's receiving and the Device would in turn be logging what part of its interface is called. This could be inspected then.

MikeHart85:
The motivation is that the IBEX team would need to restart the IOC as well to restart the simulator, which would take too long when running dozens of unit tests.

> I'm basically thinking of taking a snapshot of the device at a certain point in time (all the state, including current state of the statemachine) and applying that again, possibly to a device implementation with different state handlers in the state machine (for failure cases).

That was my thought as well... it should be possible to do this at runtime, no? Create new instance (with different setup), copy attributes, replace references to new instances, old instance disappears.

Something we haven't really talked about during the redesign is how Setups would handle scenarios where I want to change how the interface behaves (so change what the T command returns for example). Do you have any thoughts on this?

[...]

PR #210, opened 2017-03-01
Decouple adapter construction and device binding

### LINKED ISSUE #207: Remove device-argument from Adapter's init
While working on #201 (which is a pre-requisite for some other issues, such as #126) I noticed that the construction of interfaces using `DeviceBuilder` is really awkward because it requires a device instance. The only reason for that is that the binding step between interface and device happens in the `__init__`-method of each adapter.

I think removing the device-argument from init and introducing an explicit binding-step which has the device as an argument will provide more control over when that is happening and decouples construction of device/interface. It's probably also a good step towards separating the server part of `Adapter` from the device interface definition part.

MichaelWedel:
This fixes #207.

With these changes, `device` is no longer an argument of `Adpater.__init__`. This makes it possible to construct an adapter without the need of having a device object available at exactly that point. As a side-effect, it is now also much easier to "re-bind" another device, e.g. to swap out the device at runtime completely seamlessly (from the adapter's point of view). This will make it much easier to implemented #98.

Also, `device` is now a "proper" member of `Adapter`, making it less awkward to access the device when writing an interface. The internal `_device` still works, but maybe we should add a recommendation to the release notes that `device` is now available and should be used instead.

The user facing documentation has been updated.

### DIFF
--- a/docs/developer_guide/writing_devices.rst
+++ b/docs/developer_guide/writing_devices.rst
@@ -242,17 +242,17 @@ According to the specifications above, the commands are defined like this:
         out_terminator = '\r\n'
 
         def get_status(self):
-            return self._device.state
+            return self.device.state
 
         def get_position(self):
-            return self._device.position
+            return self.device.position
 
         def get_target(self):
-            return self._device.target
+            return self.device.target
 
         def set_target(self, new_target):
             try:
-                self._device.target = new_target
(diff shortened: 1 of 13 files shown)


MikeHart85:
Cannot see any issues with this.... ready to merge if you are :+1:

PR #212, opened 2017-03-10
Allow switching setups at runtime

### LINKED ISSUE #98: Ability to change setups at runtime

MichaelWedel:
This fixes #98.

With these changes it is possible to switch the setup at runtime via the control server. The functionality is implemented in `Simulation`.

I expect the implementation to change somewhat when all the adapter-related issues are taken up in the next release, so this implementation is very focused on providing only this new capability and nothing else.

### DIFF
--- a/docs/release_notes/release_1_0_3.rst
+++ b/docs/release_notes/release_1_0_3.rst
@@ -20,8 +20,6 @@ New features
 
       $ lewis-control simulation switch_setup new_setup
 
-   The simulation cycles and runtime are reset.
-
  - It has been made easier to deposit devices in an external module while maintaining control over
    compatibility with the rest of the Lewis-framework. Lewis now checks for a version specification
    in each device module against the framework version before obtaining devices, adapters and
--- a/docs/release_notes/release_1_0_3.rst
+++ b/docs/release_notes/release_1_0_3.rst
@@ -8,6 +8,20 @@ This release is currently in progress.
 New features
 ------------
 
+ - The device setup can be changed at runtime through the control server. It is not possible to switch to another device, only setups of the same device can be used. To query available setups:
+
+   .. ::
+
+      $ lewis-control simulation setups
+
+   Then, to actually activate the new setup, assuming it is called ``new_setup``:
+
+   .. ::
+
+      $ lewis-control simulation switch_setup new_setup
+
+   The simulation cycles and runtime are reset.
+
  - It has been made easier to deposit devices in an external module while maintaining control over
    compatibility with the rest of the Lewis-framework. Lewis now checks for a version specification
    in each device module against the framework version before obtaining devices, adapters and
--- a/lewis/core/simulation.py
+++ b/lewis/core/simulation.py
@@ -123,8 +123,7 @@ class Simulation(object):
     def switch_setup(self, new_setup):
         """
         This method switches the setup, which means that it replaces the currently
-        simulated device with a new device, as defined by the setup. This resets
-        the timers that keep track of the simulation.
+        simulated device with a new device, as defined by the setup.
 
         If any error occurs during setup switching it is logged and re-raised.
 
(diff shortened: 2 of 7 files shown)


test/test_core_adapters.py
```diff
@@ -93,6 +93,33 @@ def test_add_adapter(self):
...
 
+    def test_set_device(self):
+        mock_adapter1 = MagicMock()
+        mock_adapter2 = MagicMock()
+
+        collection = AdapterCollection(mock_adapter1, mock_adapter2)
+        collection.device = 'foo'
+
+        self.assertEqual(mock_adapter1.device, 'foo')
+        self.assertEqual(mock_adapter1.device, 'foo')
```
MikeHart85: I think this was meant to be `mock_adapter2.device`?

lewis/core/simulation.py
```diff
@@ -110,6 +112,38 @@ def _create_control_server(self, control_server):
...
+            self._adapters.device = self._device
+        except Exception as e:
+            self.log.error(
+                'Caught an error while trying to switch setups. Setup not switched, '
+                'simulation continues: %s', e)
+            raise
+
+    def _reset_timers(self):
+        self._cycles = 0  # Number of cycles processed
+        self._runtime = 0.0  # Total simulation time processed
```
MikeHart85: Do we really want this to happen when changing setups?

These are properties of the simulation, rather than the device. Only the device has been re-created / reconfigured... Arguably, the simulation as a whole is carrying on uninterrupted, which means you could expect these counters to just continue as they were.

MikeHart85 (commented):
Very clean solution, I like it!

Just one comment on the unit test, and a philosophical question.

lewis/core/simulation.py
MichaelWedel: I have no strong feelings about this in particular since all of that is logged away properly. In the discussion attached to the issue it came up this way, but I think you might actually be right in that it's a simulation property. It depends a bit on how this feature is used, but I guess with the current limitation of switching setups within a device package it's probably okay to let the simulation continue. Makes the implementation easier as well.

MikeHart85: I don't mind much either way, but I'm leaning towards letting it continue counting unless there is a practical reason to reset it.

I can't imagine a scenario where a user would need it to reset to 0 (if nothing else they can save the current cycles/runtime values at the point of changing setups, to use as they new starting point). But I can imagine scenarios where resetting counters can lead to unexpected behaviour (for example, if something is tracking difference in cycle count periodically, and suddenly some other part of the system resets the counters, making the next difference a huge negative number).

MichaelWedel:
I addressed the changes from the review and after reading your arguments and giving it more thought I agree with your view on the simulation counters.

docs/release_notes/release_1_0_3.rst
```diff
@@ -8,6 +8,18 @@ This release is currently in progress.
 New features
 ------------
 
+ - The device setup can be changed at runtime through the control server. It is not possible to switch to another device, only setups of the same device can be used. To query available setups:
```
OwenArnold, line 11: @MichaelWedel is setups ubiquitous enough to mention here without reference to what a setup is? I've seen this in a few places.

docs/release_notes/release_1_0_3.rst
```diff
@@ -8,6 +8,18 @@ This release is currently in progress.
...
+
+   .. ::
+
+      $ lewis-control simulation setups
+
+   Then, to actually activate the new setup, assuming it is called ``new_setup``:
+
+   .. ::
+
+      $ lewis-control simulation switch_setup new_setup
```
OwenArnold, line 21: 👍

test/test_core_adapters.py
```diff
@@ -93,6 +93,33 @@ def test_add_adapter(self):
...
+
+        self.assertEqual(mock_adapter1.device, None)
+
+        mock_adapter4 = MagicMock()
+        mock_adapter4.device = 'bar'
+
+        collection.add_adapter(mock_adapter4)
+
+        self.assertEqual(mock_adapter1.device, 'bar')
+
```
OwenArnold, line 122: Nice readable test additions!

PR #213, opened 2017-03-14
Filter Members Exposed via Control Interface

### LINKED ISSUE #194: More selective remote exposure of attributes/methods
#191 addressed a few issues that became apparent after we started using `ControlClient` in other contexts. `ControlServer` makes very few assumptions on which parts to expose and which ones not, which I think is okay because that way it stays quite simple and general. As pointed out by @MikeHart85 via Slack, this does lead to some annoyances when using `lewis-control`, particularly with non JSON-serializable members.

Since we're using JSON-RPC as a transport it's probably fair to assume that only JSON-serializable members should be exposed - all others are as of now unusable anyway. This should already take care of quite a few "annoying" cases (such as the `log` member which was pointed out specifically). This partially reverses one aspect of #191, each attribute would have to be inspected once on the server side to determine whether it should be exposed or not. I think that is acceptable, the majority of property getters should not have side effects.

When it comes to methods, it gets a bit more tricky, because there it's more of a semantic question of what is allowed and what isn't. We're already not exposing members that start with `_`, because they are considered private and that makes sense. But the solution can't be to make everything private just so it's not exposed via `ControlServer` (that might not even be active).

`ExposedObject` already has the `expose` and `exclude` parameters, which give some control over the parts of an object that should be exposed and `Simulation` already uses this. I think this is the right point to start looking at - making it easier to `exclude` stuff. Maybe `exclude` could even accept types and exclude the entire interface of the type - that would work for `CanProcess` and/or `CanProcessComposite`.

MikeHart85:
Fixes #194.

Filters members exposed by default (not overridden via the `members` param) via `ExposedObject`.

Members filtered, in addition to the previous "starting with `_`" condition, are anything defined or inherited in `Device` or `StateMachineDevice` and specifically "`log`". 

`log` had to be done explicitly since it is added dynamically by a decorator.

Also fixed an error in `control.py`... not sure how that slipped past both of us! :laughing:

### DIFF
--- a/lewis/core/control_server.py
+++ b/lewis/core/control_server.py
@@ -37,6 +37,7 @@ from jsonrpc import JSONRPCResponseManager
 
 from .exceptions import LewisException
 from .logging import has_log
+from ..devices import Device, StateMachineDevice
 
 
 class ExposedObject(object):
@@ -78,17 +79,27 @@ class ExposedObject(object):
 
         self._add_function(':api', self.get_api)
 
-        exposed_members = members if members else self._public_members()
+        exposed_members = members or self._default_members()
 
         for method in exposed_members:
             if not exclude or method not in exclude:
                 self._add_member_wrappers(method)
 
-    def _public_members(self):
+    def _default_members(self):
         """
-        Returns a list of members that do not start with an underscore.
+        Returns a list of members that are good default candidates for being exposed.
+
+        This means public members that appear to be user-created and not part of the framework.
         """
-        return [prop for prop in dir(self._object) if not prop.startswith('_')]
+        def is_valid(prop):
+            return not any([
+                prop.startswith('_'),
+                prop == 'log',
+                prop in dir(Device),
+                prop in dir(StateMachineDevice)
+            ])
+
+        return filter(is_valid, dir(self._object))
 
     def _add_member_wrappers(self, member):
         """
@@ -98,7 +109,6 @@ class ExposedObject(object):
 
         :param member: The member of the wrapped object to expose
         """
-
         method_object = getattr(type(self._object), member, None) or getattr(self._object, member)
 
         if callable(method_object):
--- a/lewis/core/control_server.py
+++ b/lewis/core/control_server.py
@@ -82,7 +82,7 @@ class ExposedObject(object):
 
         exposed_members = members if members else self._public_members()
         exclude = list(exclude or [])
-        if not exclude_inherited:
+        if exclude_inherited:
             for base in inspect.getmro(type(obj))[1:]:
                 exclude += dir(base)
 
--- a/lewis/core/control_server.py
+++ b/lewis/core/control_server.py
@@ -33,11 +33,11 @@ from __future__ import absolute_import
 import socket
 import zmq
 import json
+import inspect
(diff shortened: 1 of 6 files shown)


lewis/scripts/control.py
```diff
@@ -48,7 +48,7 @@ def show_api(remote, object_name):
             raw_value = str(getattr(obj, prop))
             value_lines = raw_value.split('\n')
 
-            current_value = value_lines[0][40:] + (
+            current_value = value_lines[0][:40] + (
```
MichaelWedel, line 51: Good catch, really weird that we did not see it earlier. But it's not enough to just change this, now all the values are gone if the string is below 40 characters. It took me a while to see that and I still don't quite understand what's going on,

lewis/core/control_server.py
```diff
@@ -78,17 +79,27 @@ def __init__(self, obj, members=None, exclude=None):
...
+
+        This means public members that appear to be user-created and not part of the framework.
         """
-        return [prop for prop in dir(self._object) if not prop.startswith('_')]
+        def is_valid(prop):
+            return not any([
+                prop.startswith('_'),
+                prop == 'log',
+                prop in dir(Device),
+                prop in dir(StateMachineDevice)
```
MichaelWedel: I don't think getting the device bases classes and `'log'` in here is necessary, it just makes the `control_server` depend on this one very specific thing. I played around a bit and I think it can be made entirely general, what do you think about this:

```python
    def __init__(self, obj, members=None, exclude=None, expose_bases=False):
        super(ExposedObject, self).__init__()

        self._object = obj
        self._function_map = {}

        self._add_function(':api', self.get_api)

        exposed_members = members if members else self._public_members()
        exclude_final = list(exclude or []) + dir(inspect.getmro(type(obj))[1]) if not expose_bases else []

        for method in exposed_members:
            if not exclude_final or method not in exclude_final:
                self._add_member_wrappers(method)
```
This automatically catches `'log'` for all subclasses of `DeviceBase` because it's in the dict of that type. For `simulation` and `interfaces` it's necessary to list `'log'` in `Simulation` explicitly where they are exposed, but in my opinion that's the "lesser evil", because it keeps `ExposedObject` general.

MichaelWedel (commented):
I tested a bit and found that now all the values are gone from `lewis-control` (see the comment above).

Also, the unit tests for `ExposedObject` need to be updated.

lewis/scripts/control.py
MikeHart85: What you describe is the behaviour I was getting before making this change... it's exactly what it's supposed to fix (`:40` means "from beginning to index 40", `40:` means "from index 40 to end"). 

Are you sure you're using the correct version of lewis-control? 

If you have lewis installed, you have to be careful to use `$ ./lewis-control.py`, and not simply `$ lewis-control`... otherwise you're using the installed version without this change. I think this may also be how this slipped passed us in the first place?

MichaelWedel: You might be right, I just tried again on a different machine being careful to use the right script. So forget everything I wrote except for "good catch" and add another "good catch".

MikeHart85:
Great review suggestion!

Made updates accordingly, with some tweaks as discussed on slack.

PR #217, opened 2017-03-22
Make it possible to expose methods and functions as PVs

### LINKED ISSUE #211: Allow PV to expose methods in addition to properties
Like for `Cmd` in `adapters.stream`, it should be possible to expose things that are not defined on the device or the interface as properties. That might for example be getter/setter pairs or even other callables with a compatible signature. This will make it easier to create EPICS interfaces for "arbitrary things".

The binding step should still result in the same `BoundPV` so that the adapter code does not need to be modified.

MichaelWedel:
This fixes #211.

With these changes it has been made easier to expose getter/setter method pairs as PVs. Free functions and other callables can be used as well, allowing for the same functionality as `Cmd`/`Var` in `stream`.

### DIFF
--- a/docs/release_notes/release_1_0_3.rst
+++ b/docs/release_notes/release_1_0_3.rst
@@ -36,9 +36,9 @@ New features
 ------------
 
  - Writing devices with an EPICS interface has been made more convenient for cases where the device
-   does not have properties, but getter and setter methods. :class:`~lewis.adapters.epics.PV` has
-   been extended to accept a wider range of values for ``target_property`` and
-   ``meta_data_property``, for example method names:
+   does not have properties, but does have getter and setter methods.
+   :class:`~lewis.adapters.epics.PV` has been extended to accept a wider range of values for
+   ``target_property`` and ``meta_data_property``, for example method names:
 
    .. sourcecode:: Python
 
--- a/docs/release_notes/release_1_0_3.rst
+++ b/docs/release_notes/release_1_0_3.rst
@@ -35,6 +35,28 @@ special characters, such as colons, it is necessary to quote them:
 New features
 ------------
 
+ - Writing devices with an EPICS interface has been made more convenient for cases where the device
+   does not have properties, but getter and setter methods. Next to
+   :class:`~lewis.adapters.epics.PV`, there is now also :class:`~lewis.adapters.epics.MethodPV`:
+
+   .. sourcecode:: Python
+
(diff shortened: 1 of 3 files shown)


lewis/examples/epics_device/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class VerySimpleInterface(EpicsAdapter):
+    """
+    This is the EPICS interface to a quite simple device. It offers 5 PVs that expose
+    different things that are part of the device, the interface or neither.
+    """
+    pvs = {
+        'Param-Raw': PV('param', type='int', doc='The raw underlying parameter.'),
+        'Param': MethodPV(('get_param', 'set_param'), type='int'),
```
OwenArnold: 👍

docs/release_notes/release_1_0_3.rst
```diff
@@ -35,6 +35,29 @@ special characters, such as colons, it is necessary to quote them:
 New features
 ------------
 
+ - Writing devices with an EPICS interface has been made more convenient for cases where the device
+   does not have properties, but getter and setter methods. :class:`~lewis.adapters.epics.PV` has
```
OwenArnold: very minor point. Should say, "but does have getter and setter methods"

OwenArnold:
@MichaelWedel Declaring a property without a setter does not require author to set the `read_only=True` argument. Is that something we should sanitise?

Setting the `read_only` flag gives a nice output at runtime 
```
Second-Int: Driver rejects value 1
```
Failing to do so seems to give garbage

OwenArnold:
@MichaelWedel. Other than that I think the changes are good to go. 👍  For ensuring the examples got updates.

MichaelWedel:
I see what you mean. This behavior is quite old I think, but it's fixed for getter/setter methods, where this happens automatically (so manual `read_only=True` is not required). It should be possible to do the same for properties. I'll look into it quickly.

lewis/adapters/epics.py
```diff
@@ -278,6 +278,13 @@ def _get_target(self, prop, *targets):
                 None)
 
         if target is not None:
+            # If the property is an actual property and has no setter, read_only can be
+            # set to True at this point automatically.
+            target_prop = getattr(type(target), raw_getter, None)
+
+            if isinstance(target_prop, property) and target_prop.fset is None:
```
OwenArnold, line 285: 👍

PR #218, opened 2017-03-23
Implement "verification mode"

### LINKED ISSUE #199: Introduce device "sanity check"

MichaelWedel:
This fixes #199.

After discussing earlier this week, the easiest solution seemed to be an additional parameter for the `lewis` script (`-V/--verify`) which simply activates debug level logging and aborts before actually starting the simulation. I've added a bit more debug output to the device discovery process to give more information at that stage.

This PR does not implement all possible debug output - this is something we can pick up on in the next release.

### DIFF
--- a/docs/release_notes/release_1_0_3.rst
+++ b/docs/release_notes/release_1_0_3.rst
@@ -61,6 +61,9 @@ New features
    This way using devices from different sources becomes more reliable for users with different
    versions of Lewis, or hint them to update.
 
+ - A new flag ``-V/--verify`` has been added to the ``lewis``-script. When activated, it sets
+   the output level to ``debug`` and exits before actually starting the simulation. This can
+   help diagnose problems with device modules or input parameters.
 
 Bug fixes and other improvements
 --------------------------------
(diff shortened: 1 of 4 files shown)


OwenArnold:
This looks good. Works for me. Code looks neat enough. Aside from the minor point with the casing of different options, I think this is good to go.

PR #219, opened 2017-03-23
Fix broken build

MichaelWedel:
This should fix the accidentally broken build.


OwenArnold:
Works locally. Good by me.

PR #221, opened 2017-03-24
Add new flag to print adapter options

### LINKED ISSUE #220: Implement "help" for new adapter command line syntax
As pointed out by @OwenArnold, there is currently no way to learn about adapter options on the command line. There should be a way to query these options.

MichaelWedel:
This fixes #220.

Specifying the `-L` option should print the adapter's configuration. It prints the actual configuration, not the default value (although that could be added as well).

It's also automatically available via the control server as `lewis-control interface configuration`.


OwenArnold:
This pretty much works as I would expect. Docs are good. Unit tests extended.

PR #225, opened 2017-03-26
Versioning in lewis-control

### LINKED ISSUE #223: Versioning on all executables
Testing out 1.0.3. A minor thing I noted was that `lewis_control` lacks a version option, which `lewis` does support.

For diagnostics one might wish to verify that they are using compatible versions of the same tools.

OwenArnold:
Fixes #223. Adds versioning option to `lewis-control`

Code review and check that no side effects have been introduced into the control client.


docs/release_notes/index.rst
```diff
@@ -7,6 +7,7 @@ called "Plankton").
 .. toctree::
     :maxdepth: 1
 
+    release_1_1_0
```
MichaelWedel: In an attempt to avoid confusion we did not include this in the index before the actual release. See comment below.

docs/release_notes/release_1_1_0.rst
```diff
@@ -0,0 +1,12 @@
+Release 1.1.0
```
MichaelWedel, line 3: For avoiding sphinx warnings when not including this document anywhere, we added `:orphan:` at the beginning. To make sure that nobody mistakes this as a secret release they haven't seen, I usually also added something like "This release in currently in progress" or something along those lines.

MichaelWedel (changes requested):
Thanks for fixing this, works as described. Once the release notes are changed this is ready to be merged.

MichaelWedel (approved):

PR #227, opened 2017-03-29
scanf format specification for Cmd and Var

### LINKED ISSUE #216: Implement scanf for StreamAdapter's Cmd and Var
We've been talking about this from the very beginning of the project - it would be much simpler for some cases to provide `scanf`-like patterns instead of regular expressions. It turns out that the author of the `scanf`-package has made it available on PyPI and it has been made Python 3 compatible:

https://pypi.python.org/pypi/scanf

There is also a function that translates the `scanf`-pattern into a regex, that would allow us to use it relatively easily without changing the code of `Cmd` and `Var` too much.

MichaelWedel:
This fixes #216.

Thanks to the [scanf](https://github.com/joshburnett/scanf) package, it's possible to make `scanf` format specifications available for `Cmd` and `Var`.

The changes are not complete, this is just my current suggestion. It preserves everything that has been implemented so far (regular expressions for `Cmd` and `Var` by default), makes it relatively easy for an interface to use scanf patterns for commands by default and still leaves the possibility to have "one offs" within a device. It does not place any syntax constraints on the patterns and adds the possibility for even more types of patterns (that can be translated to regex)...should that ever be necessary.

If there's no major disagreement I'd go forward with adding/adjusting documentation.


lewis/examples/example_motor/__init__.py
```diff
@@ -97,11 +97,14 @@ class ExampleMotorStreamInterface(StreamAdapter):
...
+    pattern_type = fmt
+
     commands = {
-        Cmd('get_status', r'^S\?$'),
-        Cmd('get_position', r'^P\?$'),
-        Cmd('get_target', r'^T\?$'),
-        Cmd('set_target', r'^T=([-+]?[0-9]*\.?[0-9]+)$', argument_mappings=(float,)),
+        Cmd('get_status', regex(r'^S\?$')),
+        Cmd('get_position', r'^P?$'),
+        Cmd('get_target', r'^T?$'),
```
MikeHart85: Are these correct?

If it's a regex, it will now match an empty string since the `?` is not escaped.

If it's a scanf format (which `pattern_type` seems to imply), are the `^` and `$` supposed to be there?

lewis/adapters/stream.py
```diff
@@ -387,31 +409,37 @@ def __init__(self, target_member, read_pattern=None, write_pattern=None,
...
             if inspect.isdatadescriptor(getattr(type(target), self.func)):
                 getter.__doc__ = 'Getter: ' + inspect.getdoc(getattr(type(target), self.func))
 
             funcs.append(
-                Func(getter, self.read_pattern, return_mapping=self.return_mapping, doc=self.doc))
+                Func(getter, read_pattern, return_mapping=self.return_mapping, doc=self.doc))
 
         if self.write_pattern is not None:
+            write_pattern = pattern_type(self.write_pattern) if isinstance(
+                self.write_pattern, string_types) else self.read_pattern
```
MikeHart85: Should end with `... else self.write_pattern` I think?

MikeHart85 (commented):
Just had a quick look between talks, I need to think about this a bit more I think.

lewis/examples/example_motor/__init__.py
MichaelWedel: Yes, it's supposed to be a scanf format and it seems that it works like this...I haven't looked at the generated regular expression, I will do that and check what this actually transforms to. But Issuing `T?` via telnet seemed to work...

As I wrote above, it's still work in progress and just a suggestion on how it could be done.

MikeHart85: I checked the regex, it's `^T\\?$` (or `r'^T\?$'`). 

This is really unexpected, isn't it? 

In scanf format, `$` has no special meaning and should match literal `$`. `^` only has a special meaning as a not in `[^...]`, outside of that I think it should also match a literal `^`. They certainly shouldn't match beginning and end of string, as they do here.

I understand it's work in progress, just thought this would be important to point out. It's a limitation that may well affect how we should design this feature as a whole.

lewis/adapters/stream.py
```diff
@@ -163,8 +186,12 @@ def __init__(self, func, pattern, argument_mappings=None, return_mapping=None, d
             raise RuntimeError('Can not construct a Func-object from a non callable object.')
 
         self.func = func
-        self.raw_pattern = pattern
-        self.pattern = re.compile(b(pattern), 0) if pattern else None
+
+        self.raw_pattern = pattern.raw_pattern
+        self.pattern = re.compile(b(pattern.regex_pattern), 0)
```
MikeHart85: I like the helper classes for explicitly differentiating between regex and scanf, but I'm still in favour of some kind of auto-detection over a per-interface default via attribute... We discussed it before, but now that there's some code to look at and consider, I'll try to sell you the idea one more time: :car:

For the device developer, it probably won't make a huge difference either way (although I do think auto-detection is a bit more intuitive / convenient) since they have to learn use whatever tools we give them, and both options get the job done. But I think it makes a huge difference on implementation complexity (and as a result, maintainability) for us here.

Consider if, right above here, we insert these two lines:
```python
if isinstance(pattern, string_types):
    pattern = regex(pattern) if pattern.startswith('^') else fmt(pattern)
```
With this, Lewis would auto-detect between `regex` and `scanf` based on a very simple (to learn and remember) rule when `pattern` is a string type. You could switch freely within the same device without needing to explicitly specify `fmt('...')` or `regex('...')`. No need to define or know that you need to override a `pattern_type` attribute to get `fmt` by default. You would still be able to specify the format type explicitly if you wanted to, but it would almost never be needed (really only if you need a `^` at the beginning of a `fmt`).

Most importantly: ALL of the other changes in this file can go away! We just need the two format helper classes and this block in `Func.__init__`. Instead feature being split across 5 different functions in 5 different classes, it would implemented in one constructor of this one class (not counting `fmt` and `regex` in both cases). This makes understanding and maintaining how the feature works much easier, since it is much more localized.

Command definitions would look like this:
```python
    commands = {
        Cmd('implicit_regex', r'^T\?$'),
        Cmd('implicit_scanf', r'T=%f''),
        Cmd('explicit_regex', regex(r'^P\?$')),  # Should never be needed, since you should always start with ^
        Cmd('explicit_scanf', fmt('^P=%f')),  # Only ever needed if a really weird device starts commands with ^
    }
```

Would there be a disadvantage to doing it this way?

MichaelWedel: I see your point that it is more convenient to define the patterns with less stuff to type and also regarding the implementation complexity. But there are still some things I'm a bit worried about with this approach in decreasing order of "worriedness":

1. Breaking existing patterns. Regular expressions that people have written so far which do not contain a `^` at the beginning will stop working. I don't know how many there are, but it may cause some surprises. It can be documented in the release notes and now we also learned about upgrade guides, but what I find dangerous in this case is that there's no easy way to decide whether to warn users if there is no `^` at the beginning of the pattern - it might be intended or not, there's no way to tell.
2. "With great convenience comes some danger"...I'm a bit afraid that one character may be easily overlooked or forgotten. That is also true for the `pattern_type` approach to some extent, but if nothing is specified, all strings will be interpreted as regular expressions consistently.
3. Extensibility. Although this may not be too likely to happen in the nearest future, if it decides to happens anyway, it is quite likely that the implementation needs to change (or am I overlooking something?). And then I fear it might be more difficult to come up with a way of doing that while preserving the old behavior (which is the new behavior we're trying to come up with now).

I do find the "not typing stuff" tempting, but as explained above, I think it might be the slightly more dangerous route to go (even setting aside point 3, which is probably too unlikely to happen anytime soon). I'd rather accept the complexity on our side and explain the `pattern_type` to the user - it's only there this feature really needs explaining besides the `fmt`/`regex`-classes. The other changes are internal and only concern us.

Is there a third way that we're not seeing?

lewis/examples/example_motor/__init__.py
MichaelWedel: This has been addressed now for the current version of `scanf` and a PR has been opened to escape those additional characters in `scanf_compile`. We need to remember to go back and change it back when a new version of scanf is available.

lewis/adapters/stream.py
MikeHart85: Some thoughts / explorations on your first point:

1. All our code currently begins all regexes with `^`.
1. I would say that if we implicitly match (or the device developer implicitly describes) expected TCP Stream commands _not_ starting with the beginning of the string, this would be a logic error. Because if you search for a match of `L?` _anywhere_ in the incoming TCP Stream command, that will equally match incoming `L?`, `notL?` and `MRGLGLGLGL?` commands. 
The last two are probably not intended to match (and should therefore cause an error response). 
If they are intended to match, I think it should be _required_ to _explicitly_ write the `pattern` to match that (or anything `.*L?`), so that it is immediately visible that this is intended when looking at the `pattern`. It should not be assumed that this is intended or desired just because the `^` is missing.
1. Testing some things in a Python console just now I ran into a surprise: 
**`re.match` assumes the `^` behaviour by default!!** (but not `$`)
So even if you do this: `Cmd('get_status', regex(r'S\?'))`, this will match `S?` and `S?foobar` BUT NOT `foobarS?`.
1. Snooping around the IBEX code a bit, I saw that almost all of their regexes start with `^`. The only excepting is this: https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/blob/afccb520b9c8e60b21b84e5ef6cc42c66c3e8ec1/lewis_emulators/neocera_ltc21/interfaces/stream_interface.py
Which is very interesting indeed.

I agree about there perhaps being a third option to consider... need to think about this some more.

MichaelWedel: Regarding, 3. I also noticed that during experimentation...I'm not sure if that makes it better or not, because I'm not sure any more why we have `^` in all *our* regexes (probably because "we've always done it like this"). And because we would basically force people to add an extra character that is not required (although to be honest I assumed it was), but in the case of scanf, where a literal `^` at the beginning is as valid as any other character it would require `fmt(...)`. Kind of also touches point 2 from my list above.

Concerning 2, I agree with you in terms of "should be a logic error"...and from number 3 follows, that this has to be done already now - even though at least I did not know that :|

Number 4 is totally unexpected and I never saw anything of that magnitude coming :)

I think there are very good reasons to go either way and I'm not totally convinced about either yet, so let's think about a way to make this better.

In the direction of auto-detection, are there other options than using `^`? Is it enough to look for `%([fdix]{1})` to guess with a high enough probability that it's a scanf pattern and treat it as a regex otherwise? For the patterns without such identifiers it would not be relevant except for escaping reasons. That still wouldn't solve the "extensibility" issue though.

In the other direction, could there be an easier way to define a sort of default pattern type? Could it be made "more automatic"?

MikeHart85: >Number 4 is totally unexpected and I never saw anything of that magnitude coming :)

Yes, that was a bit of a real-life plot-twist moment when I found that.

Number 3 was also a bit of a curve ball. Regarding "why we have `^` in all our regexes", I think you're right... "we've just always done it like this". But also "it always made sense like this". And, in the spirit of number 2, I'm almost tempted to say we should treat a missing `^` as an error in a command regex (even if we don't default to `fmt` in that case). Because `^S\?$` and `S\?$` are in fact exactly the same regex. So we don't miss any possible regexes by requiring `^`. But it is immediately obvious what `^S\?$` does and it can be very surprising what `S\?$` does.

If I saw interface commands defined like this (with or without explicit `regex`)...
```python
    commands = {
        Cmd('foo', regex(r'^S\?$')),
        Cmd('bar', regex(r'T\?$')),
    }
```
... I think I would be a) very surprised by the missing `^` on `T?` and b) even more surprised it actually still behaves the same as `S?`. So, to comply with the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I would forbid anything _not_ starting with `^` from being considered a valid regex.

Part of me is tempted to then say "well, since something that doesn't start with `^` cannot be a regex, we're free to assume it's a scanf (as long as it's not explicitly written `regex('...')`)"... but, on the other hand, I can see how this might also lead to some surprises. Especially if the user accidentally or naively misses the `^` and writes `Cmd('bar', r'T\?$')`... assuming that `T\?$` was meant to be a scanf would also be very unexpected (though arguably easier to debug, since it wouldn't match `T?` at all and you immediately know something is wrong).

For the extensibility question, do we really want to add any more formats, ever? The main reason to add `scanf` is that is matches well with the protocol files. `regex` is the general goto option for general pattern matching. I have a hard time imagining why we would want other options. Then again, I suppose the discovery of `CmdBuilder` shows that there is interest in radically different ways of specifying command format. But that is built on top of our system and doesn't involve a string where we have to decide how to interpret it.

Regarding `%([fdix]{1}`, what about `%2d`, `%*s`, `%.3f`, etc? I think it would be considerably harder to implement reliably, and lead to much more difficult rules as to "what is a regex" and "what is a scanf". The nice thing about using just `^` and nothing else is that it is a very simple to understand and remember rule and, as long as you know it, you can see at a glance whether a string will be considered regex or scanf.

Other options we haven't talked about yet:

#### 1. Factory methods
```python
    commands = {
        Cmd.regex('foo', r'^S\?$'),
        Cmd.scanf('bar', r'T=%f'),
        Cmd.builder("get_pid", arg_sep=",", ignore=r"\r\n\s").escape("QPID?").digit().build(),
    }
```
This is essentially similar to explicitly wrapping the strings, but has a few advantages:
1. `regex` and `scanf` are under `Cmd` namespace
1. No need to import formats separately
1. No need to worry about name conflicts in global space
1. Moving them outside the braces  "unclogs" the parameter values
1. Modern editors will also give you hints to available methods this way
1. They are aligned, making it easy to "batch" change them in editors that support column edits
1. The methods are free to return instances of completely different objects transparently
1. This allows radically different construction (as shown with the `CmdBuilder` style)

The only disadvantage I see is that `Var` would be locked to the same format for both read and write. This is not necessarily a bad thing to be honest.

#### 2. Different Classes (or Factory Functions)
```python
    commands = {
        CmdRegex('foo', r'^S\?$'),
        CmdScanf('bar', r'T=%f'),
        CmdBuilder("get_pid", arg_sep=",", ignore=r"\r\n\s").escape("QPID?").digit().build(),
    }
```
Very similar to factory methods. These could be actual classes or merely capitalized functions that construct `Cmd` with different parameters. `Cmd` could still exist and default to `CmdRegex` behavior. You could "change default behavior" like this:
```python
from lewis.adapters.stream import CmdScanf as Cmd
# ...
    commands = {
        Cmd('foo', r'^S\?$'),
    }
```
The disadvantage would be losing containment under `Cmd` namespace and the advantages that come with that (code organization, encapsulation, intellisense hints, importing separately).

Also, while `CmdRegex` and `CmdScanf` have compatible signatures, `CmdBuilder` (or anything else we might want to add) works completely differently. So the alias import trick is limited. The factory method syntax being a bit stricter in that regard could be seen as an advantage.

Sorry about the huge reply... it's surprising how much can be said about such a simple feature.

MikeHart85: PS: I should also ask the IBEX team what the motivation behind `CmdBuilder` was... maybe if we had `scanf`, they wouldn't need it. Not in the office today, but I'll see if I can chat with them tomorrow... or maybe just drop them an email. But it would probably be easier to discuss in person.

MichaelWedel:
This contains updates as suggested. It's a bit less regex-centric now, but regex groups are still being used in `Func`. Maybe that could move into `PatternMatcher` as well. If you're generally okay with the way this is implemented, I'd like to go ahead and update documentation.

PR #228, opened 2017-03-31
Add cycle diagram

MikeHart85:
This PR adds the Simulation Cycle diagram that was created for the ICANS poster to the README and the Framework Details section of the documentation. Wasn't sure whether it should be in both places (or maybe somewhere else instead). Let me know what you think!

Also made some other minor changes of documentation around it. Mainly changed some instances of "cycle-based" to "cycle-driven". I used that on the poster as well, and it seems a bit more descriptive.

Also moved `chopper.fsm` into the same newly created `docs/diagrams` folder as this diagram. That file isn't actually referenced anywhere, is it?


README.rst
```diff
@@ -70,9 +70,13 @@ Lewis. The ``Device`` is model for the device behaviour and internal
 memory. A ``Device`` can be represented using a ``StateMachine``, but it
 does not have to be. A ``Device`` does not include anything specific to
 the communication protocol with the ``Device``. An ``Interface``
-provides a protocol binding to a ``Device``, such as TCP stream or EPICS. The ``Device`` and
-``Interface`` are created as part of a ``Simulation`` that provides a
-"heart beat" and other environmental aspects.
+provides bindings from and protocol ``Adapter`` to a ``Device``.
```
MichaelWedel: Did you mean "from a protocol Adapter"? Or maybe "between Adapter and Device"?

README.rst
```diff
@@ -84,10 +88,10 @@ What Can You Do With Lewis?
...
    as an EPICS IOC, a TCP listener, or on any other bespoke protocol you
    like
--  Access and control the ``Device`` while it is running via a "back
-   door"
--  Access and control the ``Simulation`` while it is running via a "back
-   door"
+-  Access and control the ``Device`` while it is running via a control server
+-  Access and control the ``Simulation`` while it is running via a control server
+-  Control server can be accessed via command-line utility, Python bindings, or
+   JSON RPC.
```
MichaelWedel, line 94: I like this wording much better.

docs/developer_guide/framework_details.rst
```diff
@@ -1,19 +1,23 @@
 Framework Details
 =================
 
-The Lewis framework is built around a cycle-based statemachine that
-drives the device simulation, and shared protocol adapters that separate
-the communication layer from the simulated device.
+The Lewis framework is built around a cycle-driven core which drives the
```
MichaelWedel: Is there another word for "drives"? It sounds slightly duplicated now. What about "...cycle-driven core which controls..."?

MichaelWedel (commented):
Built the docs locally, image is included as expected. Just two small clarifications and this can be merged.

This will make the docs a bit more friendly.

docs/developer_guide/framework_details.rst
MichaelWedel: Or "...cycle-driven core which in turn drives..."?

MikeHart85: I noticed this as well, but couldn't really think of better options. I like the "in turn" version. 

The other option would be to go for something more general like "cycle-driven core which manages ...". Or leave it as "cycle-based core which drives"... but that creates inconsistency with based vs driven. Maybe I should just switch them all back to "cycle-based"?

README.rst
MikeHart85: Yes, good catch! :+1:

docs/developer_guide/framework_details.rst
MikeHart85: I added the "in turn" for now, let me know if another solution would be better.

PR #234, opened 2017-05-04
Separate Interface from Adapter

### LINKED ISSUE #232: Separate Adapter and Interface
For achieving multiple adapters and decoupled network IO, it is necessary to separate the concepts "adapter" and "interface" also in code.

This has several implications, one of the more serious ones being that existing device-interfaces will have to inherit from a new base class.

Apart from that there are a number of smaller changes (binding `Cmd`/`PV` needs different targets). We also need to make decisions regarding who creates what. Should the factory as it is now create the `Interface` and pass it to an `Adapter`? Where does the adapter come from? One idea I've had was something along these lines:

```python
class EpicsInterface(Interface):
    adapter = EpicsAdapter

```

Any other ideas?

MichaelWedel:
This fixes #232 and is ground work for threaded network IO.

### DIFF
--- a/docs/developer_guide/writing_devices.rst
+++ b/docs/developer_guide/writing_devices.rst
@@ -216,19 +216,19 @@ Implementing the device interface
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Device interfaces are implemented by sub-classing an appropriate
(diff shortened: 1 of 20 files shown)


MikeHart85:
All looks well, merging!

PR #236, opened 2017-05-09
Place adapter in separate thread

### LINKED ISSUE #224: Run adapter in separate thread
For #195 it will be necessary to run adapters in separate threads, because of some problems we've been seeing in different scenarios where the things done in one cycle are very long for some reason. This is particularly pronounced with pcaspy.

Before implementing multiple adapters, where this multi-threading will most likely be a strict requirement, we should figure out how this is done for one adapter.

MichaelWedel:
This fixes #224.

This is a suggestion for how the threaded Adapters could work. As discussed on Slack, the old serial adapter processing has been replaced completely. `AdapterCollection` does not have a `handle`-Method any more, instead the `Simulation` simply sleeps for the appropriate time.

This still needs work on the Modbus adapter.

### DIFF
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -513,7 +513,8 @@ class EpicsAdapter(Adapter):
             self._server = SimpleServer()
             self._server.createPV(prefix=self._options.prefix,
                                   pvdb={k: v.config for k, v in self.interface.bound_pvs.items()})
-            self._driver = PropertyExposingDriver(interface=self.interface, device_lock=self.device_lock)
+            self._driver = PropertyExposingDriver(interface=self.interface,
+                                                  device_lock=self.device_lock)
             self._driver.process_pv_updates(force=True)
 
             self.log.info('Started serving PVs: %s',
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -426,35 +426,37 @@ class PropertyExposingDriver(Driver):
         return False
 
     def process_pv_updates(self, force=False):
-        dt = seconds_since(self._last_update_call or datetime.now())
-        # Updates bound parameters as needed
+        """
+        Update PV values that have changed for PVs that are due to update according to their
+        respective poll interval timers.
 
+        :param force: If True, will force updates to all PVs regardless of timers.
+        """
+        dt = seconds_since(self._last_update_call or datetime.now())
+
+        # Cache details of PVs that need to update
         updates = []
 
-        for pv, pv_object in iteritems(self._interface.bound_pvs):
-            if pv not in self._timers:
-                self._timers[pv] = 0.0
+        with self._device_lock:
+            for pv, pv_object in iteritems(self._interface.bound_pvs):
+                self._timers[pv] = self._timers.get(pv, 0.0) + dt
+                if self._timers[pv] >= pv_object.poll_interval or force:
+                    try:
+                        updates.append((pv, pv_object.value, pv_object.meta))
+                    except (AttributeError, TypeError):
+                        self.log.exception('An error occurred while updating PV %s.', pv)
+                    finally:
+                        self._timers[pv] = 0.0
 
-            self._timers[pv] += dt
-            if self._timers[pv] >= pv_object.poll_interval or force:
-                try:
-                    with self._device_lock:
-                        new_value = pv_object.value
-
-                        self.setParam(pv, new_value)
-                        self.setParamInfo(pv, pv_object.meta)
-
-                        updates.append((pv, new_value))
-                except (AttributeError, TypeError):
-                    self.log.exception('An error occurred while updating PV %s.', pv)
-                finally:
-                    self._timers[pv] = 0.0
+        for pv, value, info in updates:
+            self.setParam(pv, value)
+            self.setParamInfo(pv, info)
 
         self.updatePVs()
 
         if updates:
             self.log.info('Processed PV updates: %s',
-                          ', '.join(('{}={}'.format(pv, val) for pv, val in updates)))
+                          ', '.join(('{}={}'.format(pv, val) for pv, val, _ in updates)))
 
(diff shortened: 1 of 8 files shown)


lewis/adapters/epics.py
```diff
@@ -431,16 +433,18 @@ def process_pv_updates(self, force=False):
...
-                    self.setParam(pv, new_value)
-                    self.setParamInfo(pv, pv_object.meta)
+                    with self._device_lock:
+                        new_value = pv_object.value
 
-                    updates.append((pv, new_value))
+                        self.setParam(pv, new_value)
+                        self.setParamInfo(pv, pv_object.meta)
+
+                        updates.append((pv, new_value))
```
MikeHart85: I think the entire `for` loop should be inside of a single `with lock:` rather than repeatedly locking inside the loop.

Two reasons for this, mainly:
- Locks are generally expensive to acquire (probably not an issue here though unless a device has thousands of PVs)
- The way it is, the context can switch back to the simulation in the middle of updating PVs (!!!)

The second one is the biggest problem. It means we could end up with a flow like this:
1. Simulaton cycle `A`
1. Update PVs `SIM:1`, `SIM:2`, and `SIM:3` to state after cycle `A`
1. Simulation cycle `B`
1. Update PVs `SIM:4` and `SIM:5` to state after cycle `A`(!)
1. Update PVs `SIM:1` and `SIM:2` to state after cycle `B`
1. Simulation cycle `C`
...etc

By putting the lock around the whole loop, we can ensure the PVs are always updated to a full, coherent simulation state.

If we want to avoid locking out the simulation while non-device-IO happens (pcaspy stuff), we could also split into two `for` loops... one inside the lock to grab the new `pv_object.value` and `pv_object.meta` for every `pv_object` that's due to update, and a second outside of the lock, which just uses the previously cached values to update pcaspy:
```python
with self._device_lock:
    for pv, pv_object in iteritems(self._interface.bound_pvs):
        # self._timers update junk
        if self._timers[pv] >= pv_object.poll_interval or force:
            try:
                updates.append((pv, pv_object.value, pv_object.meta))
            except blah blah:
                # blah

for pv, value, info in updates:
    self.setParam(pv, value)
    self.setParamInfo(pv, info)

self.updatePVs()
# etc
```

With the `write`s we don't have a choice, because they come as individual method calls. But that's OK for writes, because they also come as individual messages over the network and there was never a "single cycle write" guarantee (outside of the control server) anyway.

lewis/adapters/modbus.py
```diff
@@ -307,7 +308,9 @@ def process(self, data):
                 'Request: %s', str(['{:#04x}'.format(c) for c in request.to_bytearray()]))
 
             handler = self._get_handler(request.fcode)
-            response = handler(request)
+
+            with device_lock:
+                response = handler(request)
```
MikeHart85: Same idea as above, I think the lock should go around the request loop.

Don't see any easy way to optimize by splitting device IO from network IO here. Should be fine though.

lewis/adapters/modbus.py
```diff
@@ -546,8 +549,9 @@ def handle_close(self):
 
 @has_log
 class ModbusServer(asyncore.dispatcher):
-    def __init__(self, host, port, interface=None):
+    def __init__(self, host, port, interface, lock):
```
MikeHart85: Tiny nitpick, it would be a little better if the lock was names consistently throughout (`lock` or `device_lock` or `io_lock`, etc... just consistent everywhere).

lewis/adapters/stream.py
```diff
@@ -53,20 +53,21 @@ def found_terminator(self):
 
         self.log.debug('Got request %s', request)
 
-        try:
-            cmd = next((cmd for cmd in self.target.bound_commands if cmd.can_process(request)),
-                       None)
+        with self._stream_server.device_lock:
```
MikeHart85, line 56: No looping issues here... TCP stream commands always came one-by-one, without state-coherence guarantee.

This also matches real device behaviour as far as I understand. That's why devices like the Linkam T95 have commands like `Get T Status`, which returns all state information in one go.

lewis/core/simulation.py
```diff
@@ -47,7 +48,7 @@ class Simulation(object):
     In the simplest case, the actual time-delta between two cycles
     is passed to the simulated device so that it can update its internal
     state according to the elapsed time. It is however possible to set
-    a simulation speed, which serves as a multiplier for this time.
+    a simulation speed, which serv es as a multiplier for this time.
```
MikeHart85: Random space!

lewis/adapters/modbus.py
MikeHart85: Actually, thinking about this one, it's probably more similar to TCP stream than to EPICS and it's fine either way. Still, just to avoid locking/unlocking many times I would put the lock outside of the loop.

MikeHart85:
Made changes per my comments as discussed.

While testing, I noticed that our `nosetests` now take about 4-5 seconds instead of 0.1-0.2 seconds on my machine!!!

We should definitely figure out what's going on there before merging. Seems to hang one a few tests in particular.

MichaelWedel:
Have we missed anything more? While fixing the tests I discovered that the connect-disconnect test for `AdapterCollection` failed sporadically. This blocks now until the adapter thread has completed the startup, so there's no synchronisation problem anymore.

MikeHart85:
Does mocking sleep really have any effect?

While we were trying to figure out why the tests were running so slow, I commented out the sleep in simulation at one point to see if it was the cause... and removing it had no effect. I assumed it must be already mocked or otherwise replaced/ignored during the tests.

MichaelWedel:
Good question...I guess it's a kind of "insurance" to make sure `sleep` is never actually called even if we add tests later on. `simulation.py` is the only place I found where `sleep` is really used for something.

We could remove it if you want to, it's not really related.

MikeHart85:
Looks good, merging!

PR #237, opened 2017-05-25
Allow multiple adapters for a device

### LINKED ISSUE #195: Allow multiple adapters for one device
We've been discussing around this for a while, but with the likely implementation of `SocketDevice` as some sort of adapter, it becomes a concrete requirement to have multiple adapters for one device.

Taking into account the discussions in #173, `Simulation` shouldn't just store an iterable of adapters, it should be a specialized adapter container that can be exposed via the `ControlServer` properly. Otherwise is seems straightforward on the `Simulation` side of things.

An open issue is the command line interface of `lewis`. We should take a hard look at that, but I suggest to do that as a separate task once this is implemented.

MichaelWedel:
This fixes #195.

The `-p` option has been changed so it can be specified multiple times. Not specifying it results in the same behavior as now. For not using any adapters, `-n` (`--no-interface`) has been added.

The `epics_device` example has been modified so that this new feature is demonstrated somewhere.


lewis/core/simulation.py
```diff
@@ -75,17 +74,19 @@ class Simulation(object):
...
 
-    def __init__(self, device, adapter, device_builder=None, control_server=None):
+    def __init__(self, device, adapters=None, device_builder=None, control_server=None):
         super(Simulation, self).__init__()
 
         self._device_builder = device_builder
 
         self._device = device
-        self._adapters = AdapterCollection(adapter)
+        self._adapters = AdapterCollection(*adapters or [])
```
MikeHart85: This looks really weird since it might not be immediately obvious that `or` has higher precedence than `*`.

Maybe `AdapterCollection(*(adapters or []))` would be better for clarity?

Or an `adapters = adapters or []`, or `if adapters is None: adapters = []` at the top would work as well.

lewis/examples/epics_device/__init__.py
```diff
@@ -20,6 +20,7 @@
 from lewis.devices import Device
 
 from lewis.adapters.epics import EpicsInterface, PV
+from lewis.adapters.stream import StreamInterface, Var
```
MikeHart85, line 23: Seems a bit odd that the example called `epics_device` uses TCP stream... maybe rename it `dual_device` or something like that?

lewis/scripts/run.py
```diff
@@ -182,10 +189,11 @@ def run_simulation(argument_list=None):  # noqa: C901
             print('\n'.join(simulation_factory.get_protocols(arguments.device)))
             return
 
-        protocol, options = parse_adapter_options(arguments.adapter_options)
+        protocols = parse_adapter_options(
+            arguments.adapter_options) if not arguments.no_interface else {}
```
MikeHart85: This looks really weird the way it's split.

Also, didn't you mention it should be an error if both `-p` and `-n` are specified? I think that's a good idea... better than silently ignoring `-p`. Don't forget release notes if you add this.

MichaelWedel: True, the line break makes this look a little weird.

I wasn't really sure about what is best in terms of behavior, so I tried this. But maybe it is indeed better to raise an error if both options are specified, it's a small change.

MikeHart85:
Looks good, merging!

PR #239, opened 2017-06-01
Prepare release 1.0.4

MichaelWedel:
This PR bumps the version number in all required places and makes some other minor modifications.

### DIFF
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -136,7 +136,7 @@ class PV(object):
 
     .. sourcecode:: Python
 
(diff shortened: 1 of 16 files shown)


lewis/adapters/epics.py
```diff
@@ -136,7 +136,7 @@ class PV(object):
 
     .. sourcecode:: Python
 
-        class Interface(EpicsAdapter):
+        class Interface(EpicsInterface):
```
MichaelWedel, line 139: Nice catch!

PR #244, opened 2017-06-02
Fix switch_setup bug

### LINKED ISSUE #241: switch_setup broken by interface separation
When we cut out `Interface` from `Adapter` we removed `device` from `AdapterCollection`, so switching setups does create fresh device, but does not expose it (the old device is still exposed).

MichaelWedel:
This fixes #241.

This re-enables proper switching of setups. The new device is exposed by all adapters.


MikeHart85:
Looks great, works as expected!

PR #252, opened 2017-06-06
Use pytest to run unit tests

### LINKED ISSUE #249: Switch from nose to pytest as test runner
After some reading (and listening) I decided to give `pytest` a try as a replacement for `nose`. It worked well with the tests we currently have and looks like supporting #126 might be easier with `pytest`.

Given that we're only using `nosetests` as a test-runner to actually execute the tests written in `unittest`, and even the documentation states that "New projects should consider using Nose2, py.test, or just plain unittest/unittest2", I think this is a viable alternative.

MichaelWedel:
This fixes #249.

The command line to run all tests manually from the top level directory is now:

```
$ pytest test
```

To get coverage information equivalent to what we have in Travis:

```
$ pytest --cov=lewis.core --cov=lewis.devices test
```

`pytest` warned about some deprecated methods, such as `log.warn` instead of `log.warning` and `self.assertEquals` that we were oddly using in some places, I've fixed those as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,10 @@ python:
 
 install:
   - pip install -r requirements-dev.txt
-  - pip install coveralls
-  - pip install flake8
 
 script:
   - flake8 .
-  - nosetests -w test --with-coverage --cover-package=lewis.core,lewis.devices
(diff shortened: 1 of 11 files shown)


MikeHart85:
Looks good, works as expected; merging!

PR #253, opened 2017-06-12
Use tox for more reproducible testing

### LINKED ISSUE #250: Facilitate tox for better control over what is tested
The way we use our tests locally (executing `flake8` and `nosetests` manually) is quite different from what Travis is running and this has (at least for me) caused some missed flake8 failures when I forgot to execute it before opening a pull request. The same goes for the different Python environments, sometimes I miss some Python 2 or 3 thing depending on which virtualenv I am using.

[tox](http://tox.readthedocs.io/en/latest/index.html) seems to solve this in an interesting way, it creates empty virtual environments to install the package (and all its dependencies) and execute the tests. With this we'd automatically know about problems in our build/install as well.

There is integration between tox and Travis [here](https://github.com/ryanhiebert/tox-travis), it seems to simplify things on the Travis side a lot.

MichaelWedel:
This fixes #250.

There is now a configuration file for `tox`, providing a few different configurations and jobs for testing. I've followed the advice to move the `lewis` package itself into a separate `src` sub-directory. This has some consequences, namely that for being able to run the tests manually, one needs to install Lewis in the development virtualenv, usually in editable mode:

```
(lewis-dev)$ pip install -e ".[dev]"
```

The documentation has been updated accordingly and the process of running the tests manually and through tox is explained there.

To catch documentation errors earlier, building the documentation is now part of the Travis configuration as well (as a minor detail, the `imgmath` dependency for the documentation was removed because it was only used to display the Delta symbol - there was one unicode instance of it before, now all instances are unicode).

Please check also that the Dockerfile and .dockerignore files are okay.


.dockerignore
```diff
@@ -1,3 +1,6 @@
 *.pyc
 .idea
 .git
+.tox
+docs/
+.cache
```
MikeHart85: This file is for things to skip and not put inside the image when `docker build` is executing `COPY . /lewis`. Looking at this file and `.gitignore`, they are quite similar but have subtle differences.

It makes sense for this file to additionally have `.git` because the image doesn't need the whole repo. But I can't think of a reason for the other differences:
- `.coverage` in gitignore but not here
- `docs/_build` vs `docs/`
- `src/lewis.egg-info` in gitignore but not here

Would it make sense to get rid of these differences to make the ignore files more uniform?

Dockerfile
```diff
@@ -1,17 +1,11 @@
...
-# Copying this separately from the rest of lewis allows the
-# pip install step to be cached until the requirements change.
-# Rebuilding is faster and doesn't require an internet connection.
-COPY requirements.txt /lewis/
+# Lewis source is pulled in from current directory (note .dockerignore filters out some files)
+COPY . /lewis
 
 # Install any missing PIP requirements, clear PIP cache, and delete all compiled Python objects
-RUN pip install -r lewis/requirements.txt && \
+RUN pip install -e /lewis && \
```
MikeHart85: Unfortunately, moving `COPY . /lewis` up above this in order to make `pip install -e /lewis` work also means breaking the...
>Rebuilding is faster and doesn't require an internet connection.

... "feature" mentioned in the comment above.

What happened before was that docker cached these two steps:
```dockerfile
COPY requirements.txt /lewis/
RUN pip install -r lewis/requirements.txt && \
    ...
```
... when you ran `docker build` for the first time. After that, if you ran `docker build` again, it just used the cached results and didn't re-execute them unless `requirements.txt` changed. Since `COPY . /lewis` came after, a change to the source wouldn't invalidate that cache. Now, any change to the source also means all the requirements need to be downloaded and installed again.

But I think that's not too tragic / a price worth paying for the benefits of using `pip install -e`?

requirements-dev.txt
```diff
@@ -10,4 +10,5 @@ scanf>=1.4.1
 pytest
 pytest-cov
 coverage
-flake8
\ No newline at end of file
+flake8
+tox
```
MikeHart85, line 14: A quick grep seems to show that both `requirements.txt` and `requirements-dev.txt` are almost completely unreferenced now.

The only place either is mentioned is in old release notes and in `README.rst`, which suggests `pip install -r requirements-dev.txt` as part of generating the sphinx docs.

Would it make sense to remove that reference from `README.rst` and then remove the `requirements` files completely? That way we could avoid duplication / opportunity for mismatch between them and `setup.py`.

MikeHart85 (commented):
Just some thoughts and ideas on docker and requirements stuff.

No changes strictly necessary, everything looks to be working as-is.

.dockerignore
MichaelWedel: Good point about the differences.

I guess the point with `docs/_build` vs. `docs` is that we can't really exclude `docs` from git (for obvious reasons), whereas `docs/_build` is mentioned in `README.rst` under the instructions on how to build the docs (and because that's what I do on my machine).

Dockerfile
MichaelWedel: Thanks for clarifying this, I wasn't entirely sure about the implications of this. The slower build time doesn't seem like an enormous price to pay, but the "doesn't require an internet connection" sounds like it's introducing a slight inconvenience.

What about moving those into `lewis-depends`?

In a minor note, the `-e` is probably not even required, it could just be a normal install.

MichaelWedel: Another option would be to leave stuff like it was before and still do `pip install .` at the end for consistency with "normal installation".

requirements-dev.txt
MichaelWedel: I've thought about this and this morning I had the idea of just gathering the requirements in `setup.py` from `requirements.txt` (and `-dev`). Of course other people have thought about it.

Apparently both have slightly different "features", in that `requirements.txt` is essentially a list of parameters to pass on to `pip`, so there's a few things that you can do there but not in `setup.py`.

In the end I decided that it might be a good idea to keep them around, if nothing else then for tools that want to install requirements without installing the package itself (IDE, or docker, see above?)

.dockerignore
```diff
@@ -1,6 +1,11 @@
+.git
 *.pyc
 .idea
-.git
+.coverage
 .tox
+.cache
+dist/
+build/
 docs/
```
MikeHart85: You're right that gitignore cannot exclude `docs/` / has to exclude `docs/_build`, but couldn't/shouldn't dockerignore do the same to include the rest of `docs/`?

I guess it depends on whether we want to consider the docker image purely as a packaged executable, or as snapshot of `lewis` in a ready-to-run environment.

Dockerfile
```diff
@@ -1,11 +1,20 @@
...
+RUN pip install -r /lewis/requirements.txt && \
+    rm -rf /root/.cache/pip/* && \
+    find \( -name '*.pyc' -o -name '*.pyo' \) -exec rm -rf '{}' +
+
 # Lewis source is pulled in from current directory (note .dockerignore filters out some files)
 COPY . /lewis
 
-# Install any missing PIP requirements, clear PIP cache, and delete all compiled Python objects
+# Install lewis in-place and remove compiled Python objects
 RUN pip install -e /lewis && \
```
MikeHart85: I definitely like `pip install -e /lewis` better than a normal install.

This way, you can `docker exec` into a running container, or just start with `docker run -it --entrypoint /init.sh lewis`, and edit/debug/execute lewis in its own little sandbox environment that has all the dependencies pre-installed. This is particularly nice if you can't or don't want to deal with installing EPICS on the host OS but want to make or test some quick changes to lewis.

The main advantage I saw to using `pip install -e` over the old method was that it got rid of the need for `requirements.txt` to exist, which meant that the dependencies would only be listed in one place: the setup file. But if we're keeping `requirements.txt` anyway, is there much point to using `pip install -e` inside the container? 

I can think of only two effects it has:
- You can now execute `$ lewis` and `$ lewis-control` directly, instead of `$ /lewis/lewis.py` and `$ /lewis/lewis-control.py` (hidden most of the time since it's the entrypoint)
- `setup.py` may install something that `requirements.txt` didn't include (which would technically be a bug since they should be the same?)

The difference seems quite minimal. The first point won't be noticeable when just starting a lewis simulation normally. The second point is probably a disadvantage, since it could hide bugs.

If I remember right, the main reason for `lewis-depends` was that some dependencies (pcaspy and pyzmq) required build tools and various other `-dev` packages in order to install. This is pretty heavyweight and time-consuming to download and install, so we didn't want it to happen every time the `lewis` image is rebuilt. Right now, `lewis-depends` explicitly installs just those two packages and isn't aware of `requirements.txt` or any part of lewis really.

We could change that and download and `pip install -r` the latest `requirements.txt` from the master branch of `lewis`... but this means `lewis-depends` would have to be rebuilt every time `requirements.txt` changes. Or, at least, whenever we want to stop having to download/install recently added `requirements` every time `lewis` is rebuilt.

.dockerignore
MichaelWedel: True, I thought of it mostly as a packaged executable, so I wouldn't expect people to go inside the container and build the docs. Do you think that may happen?

Dockerfile
MikeHart85: Not sure what the best way to resolve this is, but I'm leaning towards just leaving it the way it was before (since we're keeping `requirements.txt` anyway) to avoid complications from getting `setup.py` involved.

MichaelWedel: I don't have strong feelings about this, but from your explanations it seems that putting the dependencies into `lewis-depends` makes things overly complicated with the constant rebuilding and all.

I'm not sure on the other hand whether involving `setup.py` makes things more complicated, but not using it would force us to bring back `lewis.py` and `lewis-control.py`.

.dockerignore
MikeHart85: Seems pretty unlikely... but perhaps you might want access to the docs in a scenario where you're debugging something on-site, with lewis installed only as a docker image, and on an isolated network that isn't connected to the internet.

It's a pretty specific / unlikely scenario, of course, but on the other hand I don't see much downside to including the contents of /docs... plus it has the upside of keeping the two ignore files more consistent with each other.

MichaelWedel: What about `test`?

Dockerfile
MikeHart85: Ah, that's right, I forgot they're completely gone now.

In that case it might be better to use `pip install -e` just for the `lewis` and `lewis-control` "shortcuts".

Alternatively, you can execute folders containing a `__main__.py` like this:
```
~/Dev/lewis$ python src/lewis
```
But on that note I think our `__main__.py` has a mistake in it... shouldn't this:
```python
from lewis.scripts.run import run_simulation
```
... be this instead...
```python
from scripts.run import run_simulation
```
... since we're referencing `src/lewis/scripts/run.py` from `src/lewis/__main__.py`?

MichaelWedel: I didn't know `__main__.py` did that as well, I was only aware of this:

```
$ python -m lewis
```

Regarding the import - we're using absolute imports throughout the entire library if I remember correctly. I don't remember if there was a particular reason for that. With `pip install -e` and so on, that's not a problem, but we could look into that issue separately.

MikeHart85: I see... only reason I brought up the imports is because `$ python src/lewis` throws an exception on that import (without lewis installed as a package).

Anyway, I'm happy to keep `pip install -e`... but then we should probably also have a `rm -rf /root/.cache/pip/* &&` step after it, same as after the `pip install -r`, just in case pip decides to stuff something in there.

MichaelWedel: What happens if you `cd` into `src` first? I think it should work in that case (`.` is added to the path implicitly).

Okay, I wasn't sure whether that step was necessary, I only looked at the image size. I'll take a look at the directory and if there's anything in it I'll add the cleanup line.

.dockerignore
MichaelWedel: The two files are now identical except for `.git` in `.dockerignore`. It's probably the best this way.

PR #256, opened 2017-06-26
Only update PV and PV meta data on actual changes

### LINKED ISSUE #254: PVs should only be updated if value would change
The `EpicsAdapter` currently updates PVs periodically, in a specified poll interval. The update should only be performed if the value actually changed, which would also provide more useful log messages regarding these updates.

MichaelWedel:
This fixes #254.

PV and PV meta-data are now updated and logged separately, and only when their values have actually changed. Also, another bug was fixed that made PVs with a `meta_data_property` read only.

To test, you could use the `dual_device`:

```
$ lewis -k lewis.examples -p epics -r localhost:10000 dual_device
```

Now use `lewis-control` to update a few values (`upper_limit`) and check in the log that the values are updated properly. Release notes will be added once #253 has been merged (the release notes file was created in that PR).

### DIFF
--- a/src/lewis/adapters/epics.py
+++ b/src/lewis/adapters/epics.py
@@ -480,24 +480,31 @@ class PropertyExposingDriver(Driver):
                     finally:
                         self._timers[pv] = 0.0
 
-        if value_updates:
-            for pv, value in value_updates:
-                self.setParam(pv, value)
+        self._process_value_updates(value_updates)
+        self._process_meta_updates(meta_updates)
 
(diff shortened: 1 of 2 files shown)


MichaelWedel:
I've addressed the `TODO` comment by opening a PR with the changes against the `pcaspy` repository.

MichaelWedel:
The upstream PR has been merged. Once a new `pcaspy` release is out, the `_get_param_info` method should be removed.

MikeHart85:
Looks good!

PR #257, opened 2017-06-26
Improve PV binding errors

### LINKED ISSUE #255: Improve exceptions for PV binding
Currently the exceptions that are raised in the PV binding step are not very helpful, because they are missing function or method names (especially `PV._get_callable`).

The messages should make it possible to trace where the errors came from.

MichaelWedel:
This fixes #255.

To test, introduce a few errors in the `EpicsInterface` of `dual_device` (lambda with additional parameter, spelling error in property name). Before these changes you get cryptic error messages that `None` could not be found or similar, after these changes there should be proper error messages plus exception information in the debug log.

### DIFF
--- a/lewis/adapters/epics.py
+++ b/lewis/adapters/epics.py
@@ -321,7 +321,7 @@ class PV(object):
             raise RuntimeError(
                 'The function \'{}\' does not look like a getter function. A valid getter '
                 'function has no arguments that do not have a default. The self-argument of '
-                'methods does not count towards that number.')
+                'methods does not count towards that number.'.format(final_callable.__name__))
 
         @wraps(final_callable)
         def getter(obj):
@@ -348,7 +348,7 @@ class PV(object):
             raise RuntimeError(
                 'The function \'{}\' does not look like a setter function. A valid setter '
                 'function has exactly one argument without a default. The self-argument of '
-                'methods does not count towards that number.')
+                'methods does not count towards that number.'.format(func.__name__))
 
         def setter(obj, value):
             func(value)
@@ -366,13 +366,14 @@ class PV(object):
         :return: Callable.
         """
         if not callable(func):
+            func_name = func
             func = next((getattr(obj, func, None) for obj in targets if func in dir(obj)),
                         None)
 
-        if not func:
-            raise AttributeError(
-                'No method with the name \'{}\' could be found on any of the target objects '
-                '(device, interface). Please check the spelling.'.format(func))
+            if not func:
+                raise AttributeError(
+                    'No method with the name \'{}\' could be found on any of the target objects '
+                    '(device, interface). Please check the spelling.'.format(func_name))
 
         return func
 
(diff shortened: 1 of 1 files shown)


lewis/adapters/epics.py
```diff
@@ -321,7 +321,7 @@ def _create_getter(self, func, *targets):
             raise RuntimeError(
                 'The function \'{}\' does not look like a getter function. A valid getter '
                 'function has no arguments that do not have a default. The self-argument of '
-                'methods does not count towards that number.')
+                'methods does not count towards that number.'.format(final_callable.__name__))
```
MikeHart85, line 324: Potentially re-introduces the same issue as was fixed in #229: 
`__call__`able `class`es and `functools.partial`s have no `__name__`.

Except `if not self._function_has_n_args(final_callable, 0):` right above this would already throw an exception, since it tries to use `inspect.getargspec` on `final_callable`, which also doesn't work with classes and partials.

Do we ever expect / want to support the `func` parameter of `_create_getter` or `_create_setter` to be a callable class or partial? 

If not, it should be fine to leave everything as-is. 
If yes, we need to fix `_function_has_n_args()` and replace `.__name__` with something that will also work for those cases (perhaps a special `utils.get_callable_name()` function, since this already came up in #229).

MichaelWedel: True, good thing you remembered that. For the time being I don't see an urgent need to support partials and/or callable classes...I don't know about "ever". Does it work for stream?

MikeHart85: No, it doesn't work with stream either... modifying `dual_device` with:

```python
from functools import partial

def multiply(x, y):
    return x * y

class VerySimpleDevice(Device):
    param = 10
    _second = 2.0

    dbl = partial(multiply, 2)
    #def dbl(self, y):
    #    return 2 * y
```
and
```python
    commands = {
        Var('param', read_pattern=r'P\?$', write_pattern=r'P=(\d+)', argument_mappings=(int,),
            doc='An integer parameter.'),
        Cmd('dbl', scanf('D=%d'))
    }
```
Works with the commented out `def` version, but fails with the partial version due to this in `Func.__init__`:
```python
        try:
            inspect.getcallargs(func, *[None] * self.matcher.arg_count)
        except TypeError:
            raise RuntimeError(
                'The number of arguments for function \'{}\' matched by pattern '
                '\'{}\' is not compatible with number of defined '
                'groups in pattern ({}).'.format(
                    getattr(func, '__name__', repr(func)), self.matcher.pattern,
                    self.matcher.arg_count
                ))
```
`getcallargs` throws an exception stating `func` is not a Python function, then our RuntimeError about argument number mismatch is displayed (since it was a TypeError).

This is certainly not a common use-case, but it's one of those things you might expect to work and be very disappointed when it doesn't. It's Python, after all, and with duck typing you'd expect to be able to use anything that's callable.

That said, there is a simple workaround: Just wrap your class call or partial in a real method or function.

MikeHart85: I'm happy to merge as-is, if you are!

OwenArnold: I was intending to take a look at this tomorrow, but happy for you to go ahead now if you are both happy.

PR #259, opened 2017-07-12
Update user guide with interesting docker use case

### LINKED ISSUE #258: Update documentation regarding docker usage
This is from a conversation with John Sparger and I really liked his suggestion for how to use the Lewis docker image with external devices.

His idea was that if you have the devices package (`-k`) on the host, mounting the argument to `-a` as a volume into makes it possible to use them. This can be useful when you don't want to install EPICS on the host.

This should go into the docker usage section.

MichaelWedel:
This fixes #258.

I've tested this locally by copying one of the example devices to another directory on the host and then following the newly added instructions.

As written in the ticket, credit for the idea goes to John Sparger.

### DIFF
--- a/docs/user_guide/usage_with_docker.rst
+++ b/docs/user_guide/usage_with_docker.rst
@@ -52,3 +52,23 @@ simulation cycle. To run a simulation 10 times faster:
 
 Details about parameters for the various adapters, and differences
 between OSes are covered in the "Adapter Specifics" sections.
+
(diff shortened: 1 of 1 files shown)


MikeHart85:
Looks good, ready to merge?

As a sidenote, this reminds me how much I don't like the way `-a`/`-k` works. I've made a separate issue for that: #260.

PR #262, opened 2017-08-21
Stream timeout mode

### LINKED ISSUE #261: Stream Protocols Without Terminators
Issue raised by @samueljackson92.

Some devices, such as the Eurotherm 2000, use a TCP Stream protocol that does not rely on terminators, or uses different terminators for different commands. This is currently not supported by our StreamAdapter, which relies on terminators to separate commands.

A preliminary look at [protocol file documentation](http://epics.web.psi.ch/software/streamdevice/doc/protocol.html#var) seems to indicate that a timeout can be used to separate commands, instead of a terminator.

> ReadTimeout = 100;
>     Integer. Affects `in` commands.
>     The device may send input in pieces (e.g. bytes). When it stops sending, how many milliseconds to wait for more input bytes before giving up? **If InTerminator = "", a read timeout is not an error but a valid input termination.**

The direction of data flow is reversed, since this is written from the IOC's perspective, but adding the same sort of timeout feature in the stream adapter should allow us to support these kinds of devices while leaving current behaviour largely intact.

MikeHart85:
Fixes #261.

Adds a `readtimeout` attribute to `StreamInterface`. A ReadTimeout is triggered when the adapter has begun receiving data (IE, a partial command), but has not yet received an InTerminator, and no further data has arrived for the number of milliseconds specified by the `readtimeout` attribute.

This behaviour is based on the description of ReadTimeout in the [Protocol File documentation](http://epics.web.psi.ch/software/streamdevice/doc/protocol.html#var), which seems to indicate that a timeout can be used to separate commands instead of a terminator.

> ReadTimeout = 100;
>     Integer. Affects `in` commands.
>     The device may send input in pieces (e.g. bytes). When it stops sending, how many milliseconds to wait for more input bytes before giving up? **If InTerminator = "", a read timeout is not an error but a valid input termination.**

This ReadTimeout is the inverse of what is implemented here, since this is written from the IOC's perspective, but it seems reasonable that it should work the same way in both directions.

## Testing

Run the supplied sample device:
```
$ lewis -k lewis.examples timeout_device
```

To interact with this device, you must switch telnet into char mode, or use netcat with special tty settings:
```
$ telnet host port
^]
telnet> mode char
[type command and wait]

$ stty -icanon && nc host port
hello world!
foobar!
```

The following commands are available:
 - `hello ` (with space): Reply with "world!"
 - `foo`: Reply with "bar!"
 - `P`: Returns the device parameter
 - `P=%d`: Set parameter to specified value

ReadTimeout is set to 2.5 seconds here.

### DIFF
--- a/src/lewis/adapters/stream.py
+++ b/src/lewis/adapters/stream.py
@@ -52,18 +52,37 @@ class StreamHandler(asynchat.async_chat):
             self._readtimer += msec
 
         if self._readtimer >= self._readtimeout:
-            self.found_terminator()
+            if not self.get_terminator():
+                # If no terminator is set, this timeout is the terminator
+                self.found_terminator()
+            else:
+                self._readtimer = 0
+                request = self._get_request()
+                with self._stream_server.device_lock:
+                    error = RuntimeError("ReadTimeout while waiting for command terminator.")
+                    reply = self._target.handle_error(request, error)
+                    self.log.debug('Error while processing request', exc_info=error)
+                self._send_reply(reply)
 
     def collect_incoming_data(self, data):
         self._buffer.append(data)
         self._readtimer = 0
 
-    def found_terminator(self):
+    def _get_request(self):
         request = b''.join(self._buffer)
         self._buffer = []
+        self.log.debug('Got request %s', request)
+        return request
+
+    def _send_reply(self, reply):
+        if reply is not None:
+            self.log.debug('Sending reply %s', reply)
+            self.push(b(reply + self._target.out_terminator))
+
+    def found_terminator(self):
         self._readtimer = 0
 
-        self.log.debug('Got request %s', request)
+        request = self._get_request()
 
         with self._stream_server.device_lock:
             try:
@@ -82,9 +101,7 @@ class StreamHandler(asynchat.async_chat):
                 reply = self._target.handle_error(request, error)
                 self.log.debug('Error while processing request', exc_info=error)
 
-        if reply is not None:
-            self.log.debug('Sending reply %s', reply)
-            self.push(b(reply + self._target.out_terminator))
+        self._send_reply(reply)
 
     def handle_close(self):
         self.log.info('Closing connection to client %s:%s', *self.socket.getpeername())
--- a/src/lewis/adapters/stream.py
+++ b/src/lewis/adapters/stream.py
@@ -37,26 +37,37 @@ class StreamHandler(asynchat.async_chat):
     def __init__(self, sock, target, stream_server):
         asynchat.async_chat.__init__(self, sock=sock)
         self.set_terminator(b(target.in_terminator))
-        self.target = target
-        self.buffer = []
+        self._readtimeout = target.readtimeout
+        self._readtimer = 0
+        self._target = target
+        self._buffer = []
 
         self._stream_server = stream_server
 
         self._set_logging_context(target)
         self.log.info('Client connected from %s:%s', *sock.getpeername())
 
+    def process(self, msec):
+        if len(self._buffer) > 0:
+            self._readtimer += msec
+
+        if self._readtimer >= self._readtimeout:
+            self.found_terminator()
+
     def collect_incoming_data(self, data):
-        self.buffer.append(data)
+        self._buffer.append(data)
+        self._readtimer = 0
 
     def found_terminator(self):
-        request = b''.join(self.buffer)
-        self.buffer = []
+        request = b''.join(self._buffer)
+        self._buffer = []
+        self._readtimer = 0
 
         self.log.debug('Got request %s', request)
 
         with self._stream_server.device_lock:
             try:
-                cmd = next((cmd for cmd in self.target.bound_commands if cmd.can_process(request)),
+                cmd = next((cmd for cmd in self._target.bound_commands if cmd.can_process(request)),
                            None)
 
                 if cmd is None:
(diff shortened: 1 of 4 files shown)


src/lewis/adapters/stream.py
```diff
@@ -37,27 +37,60 @@ class StreamHandler(asynchat.async_chat):
...
+        self._target = target
+        self._buffer = []
 
         self._stream_server = stream_server
 
         self._set_logging_context(target)
         self.log.info('Client connected from %s:%s', *sock.getpeername())
 
+    def process(self, msec):
+        if len(self._buffer) > 0:
```
MichaelWedel: It took me a while to understand that the read timeout can only happen when there's already data in the buffer. What about something like this?

```
if not self._buffer:
    return

self._readtimer += msec
# ... and so on
```

src/lewis/adapters/stream.py
```diff
@@ -633,12 +668,18 @@ class StreamInterface(InterfaceBase):
...
+
+    This class has the following attributes which may be overridden by subclasses:
 
+     - protocol: What this interface is called for purposes of the -p commandline option.
+       Defaults to "stream".
      - in_terminator, out_terminator: These define how lines are terminated when transferred
        to and from the device respectively. They are stripped/added automatically.
-       The default is ``\\r``.
+       Inverse of protocol file InTerminator and OutTerminator. The default is ``\\r``.
+     - readtimeout: How many msec to wait for additional data between packets, once transmission
```
MichaelWedel, line 683: But then again, it was mentioned in this comment, should have read more carefully. I really like the sentence from the PR description that says "readtimeout can be considered the terminator if no `in_terminator` has been specified" - that would really make the reason behind this parameter easier to understand.

Related to that, would it be more explicit to require the `in_terminator` set to `None` in that case, or does that break the workings of `asynchat`?

MichaelWedel (commented):
I'm surprised how simple this approach looks, when you mentioned the "no terminator" devices the other day I expected something much more complicated.

Could you add this to the release notes? I'll try to find some time to do the manual test you suggested, but I can't promise anything. Maybe it would be quicker if somebody else verified this.

src/lewis/adapters/stream.py
MikeHart85: You're right, if buffer is empty the next condition should not be True either, so nothing should happen. Making it return early would make that much clearer. :+1:

src/lewis/adapters/stream.py
MikeHart85: I remember having the same thought about `None` and looking at asynchat code and then deciding to use empty string instead... but I can't quite remember why exactly. Looking at asynchat again, `None` should be fine (`set_terminator` docstring even says so). I think it was other reasons that made me lean towards `''` over `None`:

An empty string is also what the protocol files use (at least the one Sam showed me that triggered this change). And, also, an empty string could not mean anything different than "no terminator". So I don't think there's anything wrong with recommending it, or any reason to require None instead.

MikeHart85:
Added to release notes.

Yes, implementing this wasn't nearly as bad as I thought it would be, especially since asynchat typically depends on terminators... for a while I thought I might have to rewrite the adapter with a different networking library.

src/lewis/adapters/stream.py
```diff
@@ -37,27 +37,63 @@ class StreamHandler(asynchat.async_chat):
...
         self._stream_server = stream_server
 
         self._set_logging_context(target)
         self.log.info('Client connected from %s:%s', *sock.getpeername())
 
+    def process(self, msec):
+        if not self._buffer:
+            return
+
+        if self._readtimer >= self._readtimeout and self._readtimeout != 0:
```
MikeHart85, line 54: By the way, PyCharm highlights this line and tells me to "Simplify chained comparison"...

This is usually a hint to chain conditions like  like `if a > b and c < b:` into `if a > b > c:`.

But I can't figure out what it wants from me here. How do you chain `a >= b` and `b != 0`? 
`0 < self._readtimeout <= self._readtimer`?
That's not quite the same, and not nearly as clear in intention. Any ideas? Am I missing something or is PyCharm just being crazy here?

src/lewis/adapters/stream.py
```diff
@@ -37,27 +37,63 @@ class StreamHandler(asynchat.async_chat):
...
+            else:
+                self._readtimer = 0
+                request = self._get_request()
+                with self._stream_server.device_lock:
+                    error = RuntimeError("ReadTimeout while waiting for command terminator.")
+                    reply = self._handle_error(request, error)
+                self._send_reply(reply)
+
+        if self._buffer:
+            self._readtimer += msec
```
MikeHart85, line 67: After thinking about it for a bit, I moved this down to the bottom with another check against `_buffer`.

This is because, when it is at the top, we could trigger a timeout on the same cycle as we first received data (if `readtimeout <= 10 ms`. This wouldn't be good, we should wait at least one cycle. `_buffer` needs to be checked again because it might have been consumed by the code above.

MichaelWedel:
I've had some time to do the manual tests you suggested, everything works as advertised. This opens up Lewis for a whole new set of typing exercise devices when the timeout is short enough ;)

I'd be happy to merge this if you think it's ready.

MikeHart85:
It's probably best to wait for @samueljackson92 to confirm that this does what he needs, and that it's now possible to implement the Eurotherm 2000 emulator he was trying to create.

Apart from that, I think it's ready to go as well.

DominicOram:
@samueljackson92 is working on a different project now. I've tested this PR with his Eurotherm emulator and it's exactly what we needed. It would be great if this could be put into a minor release soon as we are about to upgrade our Lewis version anyway. Thank you.

PR #264, opened 2017-08-30
Move control server to separate thread

MichaelWedel:
This change moves the control server into a separate thread. This makes `lewis-control` a lot more responsive.

To test, start a device with control server enabled:

```
$ lewis -r localhost:10000 -k lewis.examples example_motor
```

In a different terminal, run:
```
$ lewis-control simulation
```
The output should appear more or less instantaneously. I had to modify the code for catching the keyboard interrupt slightly, there's a `finally` clause now that stops the simulation, doing a proper shutdown of any adapters and the control server. Otherwise the control server can not exit (`_stop_commanded` never gets set).

### DIFF
--- a/src/lewis/core/control_server.py
+++ b/src/lewis/core/control_server.py
@@ -304,6 +304,7 @@ class ControlServer(object):
         if self._socket is None:
             context = zmq.Context()
             self._socket = context.socket(zmq.REP)
+            self._socket.setsockopt(zmq.RCVTIMEO, 100)
             self._socket.bind('tcp://{0}:{1}'.format(self.host, self.port))
 
             self.log.info('Listening on %s:%s', self.host, self.port)
@@ -316,7 +317,7 @@ class ControlServer(object):
                                    "args": [exception.args],
                                    "type": type(exception).__name__}}}
 
-    def process(self):
+    def process(self, blocking=False):
         """
         Each time this method is called, the socket tries to retrieve data and passes
         it to the JSONRPCResponseManager, which in turn passes the RPC to the
@@ -329,12 +330,15 @@ class ControlServer(object):
 
         If the server has not been started yet (via :meth:`start_server`), a RuntimeError
         is raised.
+
+        :param blocking: If True, this function will block until it has received data or a timeout
+                         is triggered. Default is False to preserve behavior of prior versions.
         """
         if self._socket is None:
             raise RuntimeError('The server has not been started yet, use start_server to do so.')
 
         try:
-            request = self._socket.recv_unicode(flags=zmq.NOBLOCK)
+            request = self._socket.recv_unicode(flags=zmq.NOBLOCK if not blocking else 0)
 
             self.log.debug('Got request %s', request)
 
--- a/src/lewis/core/simulation.py
+++ b/src/lewis/core/simulation.py
@@ -24,6 +24,7 @@ an :mod:`Adapter <lewis.adapters>`).
 
 from datetime import datetime
 from time import sleep
+from threading import Thread
 
 from lewis.core.adapters import AdapterCollection
 from lewis.core.control_server import ControlServer, ExposedObject
@@ -102,6 +103,7 @@ class Simulation(object):
         # Constructing the control server must be deferred until the end,
         # because the construction is not complete at this point
         self._control_server = None  # Just initialize to None and use property setter afterwards
+        self._control_server_thread = None
         self.control_server = control_server
 
         self.log.debug(
@@ -170,8 +172,7 @@ class Simulation(object):
         self._started = True
         self._stop_commanded = False
 
-        if self._control_server is not None:
-            self._control_server.start_server()
+        self._start_control_server()
 
         self._adapters.connect()
 
@@ -185,7 +186,25 @@ class Simulation(object):
         self._running = False
         self._started = False
 
-        self._adapters.disconnect()
+        self.log.info('Simulation has ended.')
+
+    def _start_control_server(self):
+        if self._control_server is not None and self._control_server_thread is None:
+            def control_server_loop():
+                self._control_server.start_server()
+
+                while not self._stop_commanded:
+                    self._control_server.process(blocking=True)
+
+                self.log.info('Stopped processing control server commands, ending thread.')
+
+            self._control_server_thread = Thread(target=control_server_loop)
+            self._control_server_thread.start()
+
+    def _stop_control_server(self):
+        if self._control_server_thread is not None:
+            self._control_server_thread.join()
+            self._control_server_thread = None
 
     def _process_cycle(self, delta):
         """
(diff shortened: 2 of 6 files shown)


src/lewis/core/control_server.py
```diff
@@ -304,6 +304,7 @@ def start_server(self):
         if self._socket is None:
             context = zmq.Context()
             self._socket = context.socket(zmq.REP)
+            self._socket.setsockopt(zmq.RCVTIMEO, 100)
```
MikeHart85, line 307: Documentation says this does the following:

> Maximum time before a recv operation returns with EAGAIN

But what is the purpose of that and why do we need it here?

src/lewis/core/simulation.py
```diff
@@ -185,7 +186,25 @@ def start(self):
...
+                    self._control_server.process(blocking=True)
+
+                self.log.info('Stopped processing control server commands, ending thread.')
+
+            self._control_server_thread = Thread(target=control_server_loop)
+            self._control_server_thread.start()
+
+    def _stop_control_server(self):
+        if self._control_server_thread is not None:
+            self._control_server_thread.join()
```
MikeHart85: It might be good to add a timeout to this, to make sure it doesn't hang forever if something goes wrong.

src/lewis/core/control_server.py
```diff
@@ -329,12 +330,15 @@ def process(self):
...
+
+        :param blocking: If True, this function will block until it has received data or a timeout
+                         is triggered. Default is False to preserve behavior of prior versions.
         """
         if self._socket is None:
             raise RuntimeError('The server has not been started yet, use start_server to do so.')
 
         try:
-            request = self._socket.recv_unicode(flags=zmq.NOBLOCK)
+            request = self._socket.recv_unicode(flags=zmq.NOBLOCK if not blocking else 0)
```
MikeHart85, line 341: Using this test script:

```python
from lewis.core.control_client import ControlClient

client = ControlClient(host='127.0.0.1', port='10000')

print 'hello '

for i in range(1000):
    sim = client.get_object('simulation')
    del sim

print 'world'
```

And running it against some arbitrary simulation:

```
$ lewis -r localhost:10000 chopper
$ python test.py
```

Lewis often crashes with this error:
```
2017-09-20 16:29:08,683 INFO lewis.ChopperEpicsInterface.PropertyExposingDriver: Processed PV updates: State=init
2017-09-20 16:29:13,126 WARNING lewis.Simulation: Stopping simulation
2017-09-20 16:29:13,127 INFO lewis.Simulation: Stopped processing control server commands, ending thread.
2017-09-20 16:29:13,128 INFO lewis.AdapterCollection: Disconnecting device interface for protocol 'epics'
  File "/home/mikehart/Dev/lewis/env/bin/lewis", line 11, in <module>
    load_entry_point('lewis', 'console_scripts', 'lewis')()
  File "/home/mikehart/Dev/lewis/src/lewis/scripts/run.py", line 245, in run_simulation
    simulation.start()
  File "/home/mikehart/Dev/lewis/src/lewis/core/simulation.py", line 184, in start
    delta = self._process_cycle(delta)
  File "/home/mikehart/Dev/lewis/src/lewis/core/simulation.py", line 220, in _process_cycle
    self._process_simulation_cycle(delta)
  File "/home/mikehart/Dev/lewis/src/lewis/core/simulation.py", line 247, in _process_simulation_cycle
    self._control_server.process()
  File "/home/mikehart/Dev/lewis/src/lewis/core/control_server.py", line 341, in process
    request = self._socket.recv_unicode(flags=zmq.NOBLOCK if not blocking else 0)
  File "/home/mikehart/Dev/lewis/env/local/lib/python2.7/site-packages/zmq/sugar/socket.py", line 457, in recv_string
    msg = self.recv(flags=flags)
  File "zmq/backend/cython/socket.pyx", line 693, in zmq.backend.cython.socket.Socket.recv (zmq/backend/cython/socket.c:7683)
  File "zmq/backend/cython/socket.pyx", line 727, in zmq.backend.cython.socket.Socket.recv (zmq/backend/cython/socket.c:7460)
  File "zmq/backend/cython/socket.pyx", line 150, in zmq.backend.cython.socket._recv_copy (zmq/backend/cython/socket.c:2437)
  File "zmq/backend/cython/socket.pyx", line 145, in zmq.backend.cython.socket._recv_copy (zmq/backend/cython/socket.c:2344)
  File "zmq/backend/cython/checkrc.pxd", line 25, in zmq.backend.cython.checkrc._check_rc (zmq/backend/cython/socket.c:10014)
zmq.error.ZMQError: Operation cannot be accomplished in current state
```

Once, I got this error instead:

```
2017-09-20 16:42:57,261 INFO lewis.ChopperEpicsInterface.PropertyExposingDriver: Processed PV updates: State=init
Resource temporarily unavailable (src/signaler.cpp:301)
Aborted (core dumped)
```

If it doesn't crash the first time, try running test.py a couple more times. Careful, on master test.py runs **MUCH** slower! But it doesn't seem to ever crash lewis.

I first saw the error just using `lewis-control` manually, the script is just to make it easier to replicate.

MikeHart85 (changes requested):
Sorry it took me so long to look at this, many other things going on right now. Finally had a chance to test it.

I love how much faster `lewis-control` comes back... no more seeing the results come in line by line. Although, it does noticeably hang for a bit before showing the response. Do you know why that is?

I ran into a random crash while using the control client. With a script that hammers the control server, it's easy to replicate.

src/lewis/core/control_server.py
MichaelWedel: If there's no timeout, it's not possible to stop the control server, because the thread hangs forever in recv. When there's a timeout, recv will fail after the timeout, raising EAGAIN, giving exceptions (such as `KeyboardInterrupt`) a chance to get noticed.

src/lewis/core/control_server.py
MichaelWedel: I think I've found the problem, there was still a call to `process` in the main thread, I must have overlooked it. I was not able to reproduce the problem after the latest commit - could you please verify?

Also, the script basically finishes instantaneously...it's amazing. I even implemented a request counter on the server side (did not commit it though) because I could not believe it...

src/lewis/core/simulation.py
MichaelWedel: Good point. Together with the recv-timeout, this will make deadlocks on exiting very unlikely.

src/lewis/core/control_server.py
MikeHart85: Ran the script several times, then upped the loop counter to 100,000... still no crashes! Nice catch!

MikeHart85 (approved):

MikeHart85:
Looks great! 

Should also be added to release notes though. Since the control server is so much more responsive, and **many** updates are possible now during a single cycle, this can have an effect on existing control API driven testing scripts and the like.

The release notes file has moved in the master branch, since the next version will now be `1.2`, so it might be best to rebase this to a newer master as well.

PS: Could you drop by our Slack when you have time?

MikeHart85:
This is a pretty significant change, so I've moved the release note entry up to the new features section. I've also reworded it a little bit.

Looks good to merge to me, going to leave it open for a while in case you want to review my changes to the notes. 

Feel free to hit the merge button if all looks good!

PR #272, opened 2017-11-21
Added Quickstart Guide to documentation

### LINKED ISSUE #271: Documentation needs a Quickstart Guide
As discussed in #270, related to #263.

The documentation needs an entrypoint that allows new users to very quickly and simply install, run and see lewis in action. This section should be as simple as possible, aiming to give a quick cross-section of use cases and features without going into any depth.

Currently we open with multiple ways to install Lewis, followed by details of how specific adapter parameters work. This is not a good entrypoint and does not give a new user a clear picture of what Lewis is or does.

MikeHart85:
Fixes #271.

Adds a quickstart section to the beginning of the documentation.

This section shows how to install Lewis via Pip, and some things you can do with a device once it is installed.

### DIFF
--- a/docs/quickstart_guide/index.rst
+++ b/docs/quickstart_guide/index.rst
@@ -27,6 +27,7 @@ Lewis can be installed with Pip using a single command:
 
     $ pip install lewis
     $ lewis --version
+    $ lewis -h
 
 
 Run the Motor Example
@@ -60,18 +61,15 @@ For this guide, we will launch the example_motor:
 
 ::
 
-    $ lewis -r :10000 -k lewis.examples example_motor
+    $ lewis -k lewis.examples example_motor
     INFO lewis.DeviceBase: Creating device, setting up state machine
     INFO lewis.Simulation: Changed cycle delay to 0.1
     INFO lewis.Simulation: Changed speed to 1.0
     INFO lewis.Simulation: Starting simulation
-    INFO lewis.ControlServer: Listening on 0.0.0.0:10000
     INFO lewis.AdapterCollection: Connecting device interface for protocol 'stream'
     INFO lewis.ExampleMotorStreamInterface.StreamServer: Listening on 0.0.0.0:9999
 
 The example motor is a TCP Stream device and listens on port 9999 on all adapters by default.
-
-The ``-r`` parameter is optional and specifies an ``adapter:port`` pair for the "Control API" to listen on. Here it listens on port 10000 on all adapters. We will see more on that in following sections.
 
 
 Connect to Motor via Telnet
@@ -100,7 +98,13 @@ T=%f     set target
 H        stop movement
 =======  =============
 
-Note that they are case sensitive. Try entering a few commands in the Telnet session:
+You can get more details, and details on the interface of any device, by using the ``-i`` or ``--show-interface`` argument:
+
+::
+
+    $ lewis -k lewis.examples example_motor -i
+
+Note that the commands are case sensitive. Try entering a few commands in the Telnet session:
 
 ::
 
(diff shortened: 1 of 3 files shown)


docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,245 @@
...
+    INFO lewis.Simulation: Changed cycle delay to 0.1
+    INFO lewis.Simulation: Changed speed to 1.0
+    INFO lewis.Simulation: Starting simulation
+    INFO lewis.ControlServer: Listening on 0.0.0.0:10000
+    INFO lewis.AdapterCollection: Connecting device interface for protocol 'stream'
+    INFO lewis.ExampleMotorStreamInterface.StreamServer: Listening on 0.0.0.0:9999
+
+The example motor is a TCP Stream device and listens on port 9999 on all adapters by default.
+
+The ``-r`` parameter is optional and specifies an ``adapter:port`` pair for the "Control API" to listen on. Here it listens on port 10000 on all adapters. We will see more on that in following sections.
```
OwenArnold: Maybe start the sentence by saying "you can ignore the `-r` parameter at this point". and then continue to say what it refers to.

docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,245 @@
...
+
+=======  =============
+Command  Meaning
+=======  =============
+S?       get status
+P?       get position
+T?       get target
+T=%f     set target
+H        stop movement
+=======  =============
```
OwenArnold, line 99: You can also extract what these commands are? Is it overload to say how you would do that at this point? Maybe introduce the `-i` argument?

docs/quickstart_guide/index.rst
OwenArnold: "Control Service" or "Back door" rather than "Control API". I think you use "Control Service" later on.

OwenArnold: As you don't need the control server immediately. Can we delay introducing it?

docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,245 @@
...
+    $ source myenv/bin/activate
+
+Lewis can be installed with Pip using a single command:
+
+::
+
+    $ pip install lewis
+    $ lewis --version
+
+
```
OwenArnold: How about mentioning `$ lewis -h` to get lewis help options?

docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,245 @@
...
+    9.106584
+
+See `the source code <https://github.com/DMSC-Instrument-Data/lewis/blob/master/src/lewis/examples/example_motor/__init__.py>`_ of the example motor if you want to see what makes it tick.
+
+
+Connect to Motor via Control Client
+===================================
+
+In addition to the simulated TCP Stream interface, Lewis provides a so-called Control Server interface, which allows you to bypass the normal device protocol and access both device and simulation parameters directly while the simulation is running. This can be very useful to debugging and diagnostics.
+
```
OwenArnold: "Think of this as a back-door for altering your simulation at runtime without making permanent changes to the usual front door interface"  (Or something similar)?

docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,245 @@
...
+    $ lewis-control simulation
+    $ lewis-control interface
+
+See the respective sections of documentation for more details.
+
+
+Control Motor via Control API
+=============================
+
+While the CLI client is convenient for manual diagnotistics and debugging, you may find the Control API more useful for automated testing. It exposes all the same functionality available on the CLI via a Python library (In fact, that is how the CLI is implemented).
```
OwenArnold: Control API = Control Server

OwenArnold (commented):
Looks really good. I have made a few minor suggestions, but this looks like a really useful quick start guide for us.

docs/quickstart_guide/index.rst
MikeHart85: Great idea!

docs/quickstart_guide/index.rst
MikeHart85: Yeah, I should probably just remove it here.

I think my idea was to avoid having to restart Lewis later, but I end up explaining how to restart it with remote access anyway.

docs/quickstart_guide/index.rst
MikeHart85: Yes, that's a good idea. I think I'll just add it as a suggestion after the table though, rather than replacing the table with the output of `-i`, since this table is a bit more compact.

docs/quickstart_guide/index.rst
MikeHart85: I never really cared for the "back-door" terminology; it feels a bit too colloquial... so I'm avoiding it on purpose here.

Perhaps I can add something along the lines of "... to keep the main device interface clean" to the last sentence instead?

docs/quickstart_guide/index.rst
MikeHart85: I'm not sure how consistent we have been about this, but generally I think these terms mean slightly different things:

- Control Server = The ZMQ/JSONRPC server that listens on whatever you pass via `-r`
- Control Client = The `lewis-control` app that ships with `lewis` and is used to talk to the Control Server via CLI
- Control (Client) API = The Python API provided by the `lewis.core.control_client` module that can be used to write scripts that talk to the Control Server (and that the Control Client itself uses)

Here, I'm talking about the Python API, as opposed to the CLI client `lewis-control`.

OwenArnold (approved):

docs/quickstart_guide/index.rst
```diff
@@ -0,0 +1,249 @@
...
+
+In addition to the simulated TCP Stream interface, Lewis provides a so-called Control Server interface, which allows you to bypass the normal device protocol and access both device and simulation parameters directly while the simulation is running. This can be very useful for debugging and diagnostics, without having to modify the main device interface.
+
+Remote access is disabled by default and enabled only if you provide the ``-r`` argument when starting Lewis. Stop the previously launched instance of Lewis by pressing ``Ctrl-C`` and run Lewis again with the ``-r`` parameter to enable remote access like this:
+
+::
+
+    $ lewis -r localhost:10000 -k lewis.examples example_motor
+
+Lewis ships with a Control Client commandline tool that allows you to connect to it. It also has an ``-r`` argument but for the client it defaults to ``localhost:10000``, which is why it is recommended to use the same value above. Leave the Telnet session running and connected and open a third terminal.
```
nikhilbiyani: When Lewis is restarted to enable the remote access, telnet session is killed. One will have to restart the telnet. The line here indicates that the telnet session is still running.

MikeHart85: Nice catch! The instructions were to start lewis with the `-r` to begin with before, so this was left over from when you didn't need to restart it.

nikhilbiyani (approved):

PR #274, opened 2018-05-07
Remove plankton references

### LINKED ISSUE #273: Some files still reference 'Plankton'
@mattclarke pointed out that some source files still reference the old project name Plankton in the license header comment.

This should be fixed to correctly reference Lewis instead.

Looks like it's mostly Julabo files that are affected.

mattclarke:
Close #273


MikeHart85 (approved):
Looks good. Only source files with references to Plankton remaining are the release notes for v1.0.0 and the README.rst (mentioning it was the previous name), which makes sense to keep that way for now.

PR #283, opened 2020-04-27
Add support for sending unsolicited messages from a device

FreddieAkeroyd:
This is used, for example, in the simulated gamry device. It sends a "charge complete"
unsolicited message some seconds after being sent a command to "start charging".

The PR adds a unsolicited_reply() method so code like the following in the device
StreamInterface can accomplish this:

    # function to send unsolicited message when charging complete
    def charged(self):
        self.handler.unsolicited_reply("STOPPED")

    # Device commanded to start charging
    def start_charging(self):
        t = Timer(self.charging_time, self.charged)
        t.start()
        return "STARTED"

### DIFF
--- a/src/lewis/adapters/stream.py
+++ b/src/lewis/adapters/stream.py
@@ -109,7 +109,7 @@ class StreamHandler(asynchat.async_chat):
 
         self._send_reply(reply)
 
-    def unsolicitedReply(self, reply):
+    def unsolicited_reply(self, reply):
         self.log.debug('Sending unsolicited reply %s', reply)
         self.push(b(reply + self._target.out_terminator))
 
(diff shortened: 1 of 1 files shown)


DominicOram:
@mattclarke, would you like to take a look at this? Otherwise I'm happy to merge it.

src/lewis/adapters/stream.py
```diff
@@ -108,6 +109,10 @@ def found_terminator(self):
 
         self._send_reply(reply)
 
+    def unsolicitedReply(self, reply):
```
mattclarke:
```suggestion
    def unsolicited_reply(self, reply):
```

mattclarke (approved):
Other than the suggested renaming, it looks good to me

PR #291, opened 2021-02-04
Add ability to send bytes to lewis

John-Holt-Tessella:
When the stream interface is sent a byte stream in tried to encode it into a string which does not work. Instead we only encode when the input type is a string.


DominicOram:
@mattclarke are you happy for me to merge this? It maintains backwards compatibility

mattclarke:
I was planning to take a look this afternoon, but now I have seen how short it is I'll do it now...
...
...
Done!

DominicOram:
Thanks @mattclarke :) Hope all is well with you!

PR #293, opened 2021-02-23
Use bytes in unsolicited reply

DominicOram:
Does the same as https://github.com/ess-dmsc/lewis/pull/291 but with `unsolicited_reply` too.


DominicOram:
Thanks @mattclarke

PR #295, opened 2021-03-16
Fix byte str mixed

John-Holt-Tessella:
We occasionally see an issue where we get a type error mixing bytes and string. Although we can not see where the error is coming from we don't want push to fall over so:

1. catch the exception and report it. 
2. convert byte terminator to string before concaternating it


DominicOram (changes requested):
I'm not sure I like this. It feels like we're just masking a genuine issue in our emulator. Emulators should provide replies and terminators in the same type and it shouldn't be up to Lewis itself to guess what type the emulator is responding with.

mattclarke:
@DominicOram John's left now, right?
Shall we close this?

DominicOram:
He has, yes. I want to remind myself of what this was for before closing it. I'll add it to my list of things to look at.

chris-d-gregory:
I understand why you are reluctant to approve this, and I don't know what the original reason for this modification is, but I wanted to comment here as we've done exactly the same thing at our site for the following case:

- Most of the values you return from your stream interface are happily handled by the default encoding applied in the StreamHandler, so you can return them as `str`, `float`, `int` etc and let lewis cast them as strings and encode them as usual.
-  Some values though need e.g. latin-1 encoding because they are out of range for utf-8 (or whatever the system default is). So you need to return them from the interface already correctly encoded. You then set a return mapping of say `lamda x: x` so lewis doesn't try to convert it back to a string. 
- Now you have a choice: 

1. Go through and change all your return values to bytes - including those for which no special handling is needed - which might be a lot of otherwise unnecessary code. Then return an encoded out terminator too so that the two can be concatenated when building the complete reply. This works with the released version of lewis. 
2. Make the modification in this PR and you only have to manually encode those problem return values and leave the rest as they are. The fact that this modification checks the type of the out terminator and the reply separately to see what needs encoding allows this to happen. 

I accept that might not be a good enough reason to approve the PR.

DominicOram (changes requested):
On second thoughts I think this is fine. Whilst ideally people implementing emulators should be returning the correct type @chris-d-gregory is right that we can't trust them to do this and so LeWis should try and do the best it can and assuming the terminator is correct seems like the best assumption. However, it does need:
* Tests
* Documentation somewhere saying this is the assumption LeWis is making

I'm happy to do these things when I get some time

DominicOram:
I've changed it so that Lewis is as forgiving as possible now and added a test. I don't think docs are required as they'll basically boil down to "specify it in any type you want". @mattclarke and @chris-d-gregory look ok to you?

mattclarke:
👍

chris-d-gregory:
Yes, looks good.

Issue #297, opened 2021-06-15
What is the correct way to send unsolicited replies based on events ?


auphofBSF:
My use case is to simulate a basic serial device over tcp  that responds to commands but also sends unsolicited messages

The doc's strings and Readthedocs are great but I feel I am missing something 

I am using a TCP Stream Based on Motor Example

I have successfully implemented Cmd's and get the correct responses

I would like to get an `unsolicited_reply()` from the `StreamHandler`   when the following events happen

1. a new connection is made to the simulation interface with Telnet Host:Port (no data yet sent from Client)  -> Generates a Welcome message
2. a periodic (Non Regular - random time interval)  event happens in the simulated Device subclassing `StateMachineDevice` and sends an `unsolicited_reply` to the TCP client, ie simulating a state change through `lewis-control device ....`
3. a periodic (regular - fixed time interval) variant of 2.

mattclarke:
@DominicOram Any ideas/tips for this?

DominicOram:
Does https://github.com/ess-dmsc/lewis/tree/message_on_init fix point 1? You should be able to add something like the following in your strream interface:
```
def initial_message(self):
    return b"Hello World"
```
I'll think about 2 and 3 later today...

auphofBSF:
Thank you @DominicOram `initial_message` works well.  

With regard to 2&3 I have spent a few hours trying to see how find the current active `StreamHandlers` from the `process` method in `StateMachineDevice`, getting a deeper understanding of `lewis` but have not been successful yet,   so look forward to your thoughts, Thank you again

DominicOram:
Yh, the issue is that `unsolicited_reply` is a bit of a hack. The `StreamInterface` side holds a reference to the `device` but not the other way round. This means when we've used `unsolicited_reply` in the past we've ended up putting what should really be [`device` logic into the `StreamInterface`](https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/blob/master/lewis_emulators/gamry/interfaces/stream_interface.py) or we've set the unsolicited reply to just [send periodically]( https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/blob/master/lewis_emulators/mecfrf/interfaces/stream_interface.py). I guess the second link solves 3 for you but still isn't elegant.

auphofBSF:
Selfishly only considering the `adapters.stream`,  

[...]

PR #298, opened 2021-06-17
WIP  on 297 unsolicited messages, from initial connection and on device events

auphofBSF:
This PR  attempts to solve #297 , it provides  mechanisms to

1) provide an initial message on a TCP client connecting to a StreamInterface - thank you @DominicOram 
2) provides a mechanism for a device to generate a message to TCP clients such as from an event and it is sent to all `StreamHandlers` with connected clients. This could be from regular periodic events or any other non regular computed events

In addition an example  is provided in `lewis.examples` accessible as `lewis -k lewis.examples simple_eventing_state_device`

Subject to initial review, docs and possibly some tests need to be generated

### DIFF
--- a/lewis/adapters/stream.py
+++ b/lewis/adapters/stream.py
@@ -22,6 +22,7 @@ import asyncore
 import inspect
 import re
 import socket
+from collections import deque
 
 from scanf import scanf_compile
 
@@ -47,6 +48,7 @@ class StreamHandler(asynchat.async_chat):
         self._set_logging_context(target)
         self.log.info("Client connected from %s:%s", *sock.getpeername())
 
+        self._send_event_message_queue = deque([])
         initial_message = self._target.initial_message()
         if initial_message:
             self.unsolicited_reply(initial_message)
@@ -57,6 +59,11 @@ class StreamHandler(asynchat.async_chat):
             self.unsolicited_reply(event_message)
 
     def process(self, msec):
+
+        # if there are any event messages to send, send them
+        while len(self._send_event_message_queue) > 0:
+            self.send_event_message(self._send_event_message_queue.popleft())
+
         if not self._buffer:
             return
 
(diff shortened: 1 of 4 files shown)


DominicOram:
I'm on holiday now until the 28th but will hopefully have a chance to look on my return. Do you have any thoughts @mattclarke?

lewis/adapters/stream.py
```diff
@@ -178,7 +195,12 @@ def close(self):
 
     def process(self, msec):
         for handler in self._accepted_connections:
+            if self._device_event_message:
```
mattclarke, line 198: Perhaps I am missing something but where does `self._device_event_message` get set?

mattclarke:
I think this looks promising, obviously as a WIP there are a few rough edges.

I am wondering about whether we need an new abstraction for this type of thing, but I would need to remind myself of how lewis works ;)
What I am wondering is if the device should just keep track of whether it has messages to send and something else grabs them and sends them to the adapters rather than giving the device direct access to the adapters.

We'll see what @DominicOram thinks when he gets back, but I think this implementation would be okay.

DominicOram:
Sorry, I haven't forgotten about this, it's just low priority for me right now. Are you wanting it to be merged soon @auphofBSF?

auphofBSF:
No problem, I have been on other issues as well, I have been using my forked branch very effectively, but had issues when wanting to push messages from a child device.  @mattclarke comment about abstraction is valid and needs some thought. I have added some quick hacks to have messages emitted by the parent device, which got me into looking how hierarchy of devices and partitioning devices works and getting lost in that.

 I got to a good working solution, I need to tidy these up and add to this PR

As initially discussed this PR is still WIP, so do not expect it to be Merged as is.  I don't know what bandwidth I will have to abstract this messaging further. I will look to push a more refined child/parent message handler and update this PR next week and that may be as far as I can go, Thanks for keeping in touch

PR #302, opened 2021-08-10
Update Error Handling to Produce Clearer Invalid Regex Errors

JackEAllen:
Invalid Regex does not currently produce a clear, meaningful error message. 

This pull request aims to fix this by returning which function the invalid regex is present within and the invalid regex command as part of a runtime error.

### Testing Performed Before Making this Pull Request

- Install this version of lewis (`%PYDIR%\python.exe -m pip install --force-reinstall git+git://github.com/ISISComputingGroup/lewis@ba2773bc3f0e0deebd2fdcf66588836b6f3a134d`)
- Write an emulator with a bad regex in the stream interface
- Confirm that the error message produced points you at which command is wrong


mattclarke:
This looks good - would it be possible to add a test for this?

DominicOram:
@JackEAllen, I know you were struggling to write the tests due to the dependencies of `StreamHandler`. I had to deal with this in https://github.com/ess-dmsc/lewis/pull/295/files, maybe you can use some of that?

mattclarke:
Would something like this work?
```
def test_func_throws_on_construction_if_regex_invalid():
    f = lambda : 0

    with pytest.raises(RuntimeError):
        func = Func(f, "NOT A REGEX")
```
If this is possible then it can go into a new test file called `test_stream_func.py`.
If you want extra brownie points you could add tests for the other throwing parts of the constructor `if not callable(func):`, `inspect.getcallargs(func, *[None] * self.matcher.arg_count)`, etc. ;)

JackEAllen:
> Would something like this work?
> 
> ```
> def test_func_throws_on_construction_if_regex_invalid():
>     f = lambda : 0
> 
>     with pytest.raises(RuntimeError):
>         func = Func(f, "NOT A REGEX")
> ```
> 
> If this is possible then it can go into a new test file called `test_stream_func.py`.
> If you want extra brownie points you could add tests for the other throwing parts of the constructor `if not callable(func):`, `inspect.getcallargs(func, *[None] * self.matcher.arg_count)`, etc. ;)

Thank you! I always forget that lambda's are a thing as I'm usually so against using them. But they do have their use cases! 
Works perfectly for me. Will add a few more unit tests to justify placing in a new file.

DominicOram:
I think lambdas are awesome, why don't you like them?

mattclarke:
I've pulled the changes into https://github.com/ess-dmsc/lewis/pull/305 so I could tweak the formatting etc.

PR #312, opened 2024-02-22
Fix GitHub links

minijackson:
I assume this is due to a branch move from master to main.

Also maybe related: it appears the documentation links all return 404, maybe this is because readthedocs is trying to build the master branch, which doesn't exist anymore?

Thanks a lot!


mattclarke (approved):

mattclarke:
@minijackson Thanks for the corrections.

I killed the readthedocs integration as RTD were requesting that we update the security and I didn't have time to do that.
When I have some time I'll convert the docs from rst (urgh!) to markdown and see about restoring in it.

PR #316, opened 2024-10-17
Ruff pyright day

jackbdoughty:


lewis/devices/__init__.py
```diff
@@ -139,18 +139,16 @@ def _get_state_handlers(self):
...
 
         :return: The initial state of the state machine.
         """
-        raise NotImplementedError(
-            "_get_initial_state must be implemented in a StateMachineDevice."
-        )
+        raise NotImplementedError("_get_initial_state must be implemented in a StateMachineDevice.")
 
-    def _get_transition_handlers(self):
+    def _get_transition_handlers(self) -> dict[tuple[State, State], Callable]:
```
Tom-Willemsen: could be nice to specify what the signature of the callable is expected to be?

I think it's `Callable[[], bool]`

but not critical

Tom-Willemsen (approved):
Looks pretty much fine at first glance as far as I can tell, one minor suggestion, feel free to take/leave that one and then merge.

Then we'll figure out how to update the pypi package

Tom-Willemsen (approved):

PR #323, opened 2025-01-13
Correct type hint for _get_initial_state()

rerpha:


Tom-Willemsen (commented):
`_get_transition_handlers` is also type hinted incorrectly it should be `Mapping[tuple[str, str], Callable[[], bool]]:`

PR #334, opened 2025-10-30
run tests on windows

rerpha:
i accidentally pushed [this](https://github.com/ISISComputingGroup/lewis/commit/544fff75805b6f1cbfe67c82ce7ea9387c9f07f8) to master which is what _actually_ fixes the system tests on windows - it was picking up the system installed python exe on windows, `uv` actually solves this for us by using the local venv and not having to worry about running `source .venv/bin/activate || .venv\scripts\activate`

### DIFF
--- a/system_tests/lewis_tests.py
+++ b/system_tests/lewis_tests.py
@@ -1,6 +1,7 @@
 import contextlib
 import os
 import subprocess
+import sys
 import time
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


system_tests/lewis_tests.py
```diff
@@ -34,7 +34,7 @@ def julabo_simulation():
 
 
 def run_control_command(mode, command, value):
-    subprocess.check_output(["uv", "run", str(LEWIS_CONTROL_PATH), mode, command, value], env=os.environ).decode()
+    subprocess.check_output(["uv", "run", str(LEWIS_CONTROL_PATH), mode, command, value]).decode()
```
Tom-Willemsen: Wouldn't it be better to use `sys.executable` here? That gives the path to the _current_ interpreter, which will be a correct path for both windows & linux as well as not creating a hard test dependency on `uv`.

rerpha: good idea - have pushed a change to do so

Issue #335, opened 2026-03-19
Replace pcaspy with p4p


anderslindho:
For the EPICS adapter, Lewis uses pcaspy (which I believe might use legacy asyncio modules). The handling of this dependency also seems a bit odd.

We should instead use the more modern p4p, built on the reimplementation of the PVA stack.

Tom-Willemsen:
To be honest, I'd forgotten we even shipped an `[epics]` optional dependency! This functionality is not used at ISIS, and a brief search on github doesn't reveal any downstream repositories depending on `lewis[epics]`. I'm unsure if it's ever been used other than by ESS.

My first instinct would be to prefer adding a new `epicspva` optional dependency rather than reusing the `epics` dependency group for `pva` servers. This would be comparable to approaches in [`fastcs`](https://github.com/DiamondLightSource/fastcs/blob/main/pyproject.toml#L37) or [`ophyd-async`](https://github.com/bluesky/ophyd-async/blob/main/pyproject.toml#L34) which both also have this pattern of optionally supporting both EPICS CA and PVA.

anderslindho:
I would be in favour of using a more specific name (e.g. `pva` and `ca`), but I think the pcaspy usage should be removed regardless.

As I mentioned in the original report, I think the pcaspy dependency also may relevant for #320.

Tom-Willemsen:
Ok - I guess I am just being cautious as I don't know for sure whether there are any users are using the existing EPICS CA support.

From a github search, I can only find the following usages:
- https://github.com/michele-brambilla/estiasim/blob/96f9469524ae34559307f1a25810f00c067662df/lewissim/selene_motor/__init__.py#L4 - doesn't look very active.
- https://github.com/michele-brambilla/estiasim/blob/96f9469524ae34559307f1a25810f00c067662df/lewissim/spss7/__init__.py#L3 - same repo as above

Given that the only usages are those above (both seem to be related to the ESS Estia beamline), and the `ESSChopper` built into lewis, I think I am ok with the pcaspy-based `EPICSInterface` being removed if neither are needed at ESS any more.

anderslindho:
6 years ago. I think I remember that name - that they worked at PSI - but I don't know if they work with ESS anymore. I can't find their profile on our intranet (but I do find an account on our Slack instance). That was around the time that Lewis was being developed, IIRC, so maybe just early testing.

[...]

PR #336, opened 2026-06-25
Convert adapters to use asyncio instead of the deprecated asynchat/asyncore

### LINKED ISSUE #320: remove asynchat/asyncore and use asyncio instead
If we want to use python >=3.12 we need to remove this as it's been deprecated until then and completely removed in 3.12 onwards. 

it's currently used by the stream handler, #306 says a bit about this and how we might change it to suit. 

### acceptance criteria
- [ ] asynchat and asyncore are removed from use in favour of asyncio
- [ ] our python testing matrix tests against 3.12 and beyond
- [ ] the pyproject.toml is updated to use `>=3.10` as its python-versions

kovacskristof:
Convert adapters to use `asyncio` instead of the deprecated `asynchat/asyncore`

Closes #320


anderslindho:
Writing here since I was requested to have a look via e-mail.

This looks pretty nice (assuming it passes tests once the workflow is started), but I would personally like to first see test coverage for the modbus adapter, and maybe also for the "EPICS" adapter. And I think the EPICS adapter maybe has to be replaced to fully get rid of the legacy python asyncio interface - see #335.

Tom-Willemsen:
The `ruff` CI failure is due to us using [a central `ruff.toml` for all our python code](https://github.com/ISISComputingGroup/reusable-workflows/blob/main/ruff.toml). But I think to make it easier for people to contribute, I'll make this a project-local one shortly so that `lewis` stays more self-contained.

The pyright CI failure is because `pyright` doesn't quite understand the `@has_log` decorator. In other places, we forward-declare just the _type_ of log, i.e.:

```python
def __init__(...):
    self.log: logging.Logger
    ...
```

---

I will try to have a more detailed look through the changes and a functional test with some of our ISIS emulators over the coming days.

lewis/adapters/stream.py
```diff
@@ -124,61 +136,70 @@ def found_terminator(self) -> None:
...
                 reply = self._handle_error(request, error)
 
-        self._send_reply(reply)
+        await self._send_reply(reply)
 
-    def unsolicited_reply(self, reply) -> None:
+    async def unsolicited_reply(self, reply) -> None:
         self.log.debug("Sending unsolicited reply %s", reply)
-        self._push(reply)
+        await self._push(reply)
```
Tom-Willemsen: This `unsolicited_reply` function is public API which can be called directly from emulators - see [here](https://github.com/ISISComputingGroup/EPICS-DeviceEmulator/blob/f2a674ecb13fbd7d345f0264cc85a3c9e3385d97/lewis_emulators/mecfrf/interfaces/stream_interface.py#L44) for an example of an emulator that does this to emulate a device which sends a status packet once per second, unprompted.

This change from sync to async is then a breaking change for downstream emulators, as it would now need to be `await`ed. So we need to either:
- Find a way to wrap / keep the existing sync API to add an 'unsolicited message' to the queue
- Provide a different way to solve the same problem (sending unprompted messages from an emulator), and write release notes / a migration guide describing how to migrate an emulator which currently uses `unsolicited_reply` (like the one above) to the new scheme.

I think the current approach was always a bit of a hack, and we don't do this in very many downstream emulators, so I am quite open to changing the _approach_ as long as the _capability_ to send asynchronous messages remains.

lewis/core/adapters.py
```diff
@@ -318,17 +319,17 @@ def _start_server(self, adapter: Adapter) -> None:
             if not self._running[adapter.protocol].is_set():
                 raise LewisException("Adapter for '%s' failed to start!" % adapter.protocol)
 
-    def _adapter_loop(self, adapter: Adapter, dt: float) -> None:
+    async def _adapter_loop(self, adapter: Adapter, dt: float) -> None:
         adapter.device_lock = self._lock  # This ensures that the adapter is using the correct lock
-        adapter.start_server()
+        await asyncio.create_task(adapter.start_server())
```
Tom-Willemsen: Why do we need a `create_task` here if we are immediately `await`ing it?

lewis/adapters/stream.py
```diff
@@ -33,36 +31,48 @@
...
+        self._writer = writer
 
         self._set_logging_context(target)
-        self.log.info("Client connected from %s:%s", *sock.getpeername())
 
-    def process(self, msec) -> None:
+    async def handle_client(self):
+        while True:
+            try:
+                msg = await self._reader.readuntil(self._in_terminator)
```
Tom-Willemsen: This doesn't work for devices which use an empty `in_terminator`, which should treat a timeout as a valid input termination.

```
  File "C:\Instrument\dev\lewis\lewis\adapters\stream.py", line 183, in handle_accept
    await handler.handle_client()
  File "C:\Instrument\dev\lewis\lewis\adapters\stream.py", line 52, in handle_client
    msg = await self._reader.readuntil(self._in_terminator)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Instrument\Apps\Python3\Lib\asyncio\streams.py", line 610, in readuntil
    raise ValueError('Separator should be at least one-byte string')
ValueError: Separator should be at least one-byte string
```

lewis/adapters/stream.py
```diff
@@ -73,11 +83,12 @@ def collect_incoming_data(self, data) -> None:
 
     def _get_request(self):
         request = b"".join(self._buffer)
+        request = request.rstrip(self._in_terminator)
```
Tom-Willemsen: [`rstrip`](https://docs.python.org/3/library/stdtypes.html#str.rstrip) specifies a set of characters to be removed. So `rstrip("\r\n")` removes _any_ trailing `\r` or `\n` characters, not just a `"\r\n"` sequence. You may want `removesuffix` instead.

Tom-Willemsen (commented):
This looks directionally good - have left a few comments in the code about some more detailed code-level points.

lewis/adapters/stream.py
balintpeceli: We haven't realized that this was a public API, thanks for pointing this out!

We tried to strategically keep the changes at the minimum and prefer backwards-compatibility over `asyncio` "modernity". The full adapter internals could be restructured to adopt `asyncio` concepts more naturally, but we intentionally avoided that. We would prefer to tackle that later in scope of a subsequent PR if really wanted.

Following that strategy we could have a look at the [WriteTransport.write API](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.WriteTransport.write) and try to free the entire write-path from `await`ing.

Could you please give us some instructions how to properly test this behavior with the appointed emulator or any other alternatives?

lewis/core/adapters.py
balintpeceli: Although in this case it's almost the same, we propose to create a task here for the following reasons:

- this way it clearly documents what tasks we want to schedule,
- [asyncio docs](https://docs.python.org/3/howto/a-conceptual-overview-of-asyncio.html#await) warns us that "Unlike tasks, awaiting a coroutine does not hand control back to the event loop!" - which might cause troubles in the future if potentially new tasks will have to be scheduled

lewis/adapters/stream.py
balintpeceli: This use case was indeed overlooked. The fix is on the way.

lewis/adapters/stream.py
Tom-Willemsen: > We tried to strategically keep the changes at the minimum and prefer backwards-compatibility over asyncio "modernity".

Yes, I think this is the right general approach. We'd certainly prefer to avoid large changes required on the emulator side for most emulators. However I am open to changes specifically to the "unsolicited command" mechanism on the emulator side, as we only have <5 emulators which actually use this.

To test with the emulator I linked:
- Check out the [`EPICS-deviceEmulator`](https://github.com/isisComputingGroup/epics-deviceemulator) repository (mine is in `c:\instrument\apps\epics\support\deviceemulator\master`
- Run lewis with something like:
```
c:\Instrument\Apps\Python3\python.exe -m lewis -k lewis_emulators mecfrf -p "stream: {bind_address: 127.0.0.1, port: 57677}" -a c:\Instrument\Apps\EPICS\support\DeviceEmulator\master\
```
- With the emulator running, telnet to `127.0.0.1:57677`; you should get a message including the bytes `DATA` once per second, without you sending anything. The rest of the message is in a horrible binary format for that device unfortunately.

lewis/core/adapters.py
Tom-Willemsen: I am still not sure I fully understand, but can you try to write the explanation in a code comment (for future readers) and then I will think harder about whether this makes sense.

rerpha:
https://github.com/ISISComputingGroup/lewis/pull/337 now merged so you can just use `ruff format` and `ruff check` etc. as the CI does.

lewis/adapters/stream.py
kovacskristof: Tested the new version with the emulator, and it works correctly (as far as I can tell). A small independent modification was necessary in the emulator though to be able to run it. Assembling the message triggered an assert due to the fact that `msg += struct.pack("L", 0)` added 8 bytes to the message instead of 4 as the fixed `msg += struct.pack("<L", 0)` does.

lewis/adapters/stream.py
kovacskristof: Found an issue with the function as a whole. Nothing ensured that if a client would have sent two commands in one TCP packet, then both would be parsed correctly. `CMD1\r\nCMD2\r\n` would only be truncated to CMD1\r\nCMD2 and considered as one long command. The updated version parses out one command only, and leaves the rest in the buffer for the next cycle.

lewis/adapters/stream.py
Tom-Willemsen: > A small independent modification was necessary in the emulator though to be able to run it. Assembling the message triggered an assert due to the fact that msg += struct.pack("L", 0) added 8 bytes to the message instead of 4 as the fixed msg += struct.pack("<L", 0) does.

I expect this is probably just because we run on Windows usually, where a `long` is 4 bytes, as opposed to 8 bytes on 64-bit Linux which I assume you're running...

...thoroughly unrelated to this change I think :)

lewis/adapters/stream.py
```diff
@@ -124,61 +169,101 @@ def found_terminator(self) -> None:
...
-        self._push(reply)
+        if self._stream_server._loop is None:
+            raise RuntimeError("Cannot send unsolicited reply: server not started.")
+        asyncio.run_coroutine_threadsafe(self._push(reply), self._stream_server._loop).result(
+            timeout=5.0
+        )
 
-    def handle_close(self) -> None:
-        self.log.info("Closing connection to client %s:%s", *self.socket.getpeername())
+    async def handle_close(self) -> None:
```
Tom-Willemsen, line 184: Just for your info - I am seeing a local issue with IOC clients that disconnect and reconnect back to the same emulator not getting messages after they reconnect. I _think_ it might be caused by something in this new logic.

I will try to debug this in more detail locally / provide a minimal test-case shortly.

lewis/adapters/stream.py
```diff
@@ -124,61 +169,101 @@ def found_terminator(self) -> None:
...
+        if self._closing:
+            return
+        self._closing = True
+        # Setting the target.handler to None breaks client code, so
+        # rather set it to a dummy handler that does nothing.
+        self._target.handler = _NullStreamHandler()
```
Tom-Willemsen: This fixes a problem I was seeing with multiple clients - particularly with unsolicited replies.

```suggestion
        
        if self._target.handler is self:
            del self._target.handler

```

We can probably also remove the `NullStreamHandler` after we've applied this change.

kovacskristof: Done so. The tests are running and the emulator seems to work fine as well.

Tom-Willemsen: Thanks - I'll test this against a wider set of our ISIS emulators now.

Most of them look to be working as expected, think there might be a couple of edge cases I still need to debug locally.

kovacskristof: Thank you, let me know if I should modify something.

lewis/adapters/modbus.py
```diff
@@ -529,59 +531,101 @@ def _handle_write_multiple_registers(self, request):
...
+        self._closing = True
+        sock = self._writer.get_extra_info("socket")
+        if sock is not None:
+            try:
+                self.log.info("Closing connection to client %s:%s", *sock.getpeername())
+            except OSError:
+                self.log.info("Closing connection to client (peer address unavailable)")
+        if not self._writer.is_closing():
+            self._writer.close()
+        await self._writer.wait_closed()
```
Tom-Willemsen: This fixes many of the issues I was seeing locally - I think if the *remote* end closes the TCP connection (e.g. an IOC restarts), rather than the local end, we were hitting this case with a `ConnectionResetError`. It's also possible there's some platform-specific subtlety.

```suggestion
        try:
            await self._writer.wait_closed()
        except ConnectionError:
            pass
```

kovacskristof: Added this check for the stream adapter as well

Tom-Willemsen: Ah yes thanks - I actually intended to put this comment on the stream adapter and misplaced it :)

Having the same check in both locations sounds correct.

Tom-Willemsen (approved):
I intend to get this merged soon; it's just waiting on a couple of related pull requests on our end.

kovacskristof:
Thank you!

